PR #16, opened 2017-03-30
Add basic root detection for Android library

### LINKED ISSUE #3: Implement basic root detection support in Android SDK
As a nice auxiliary SDK feature, we could implement a simple jailbreak detection.

We already have this feature on iOS, we need to support it on Android as well.

TomasKypta:
The root detection uses couple of well-known techniques.
BusyBox detection was also added but is kept disabled for the time being
since it can cause false positives and an improvement of the detection is needed.

Fix #3


petrdvorak:
Nice - thank you, @TomasKypta! 👍

PR #26, opened 2017-04-20
Issues/2 jcenter deployment

hvge:
This set of changes is adding semi automatic deployment of android builds to the **jcenter** maven repository. 

Major changes:
- `ios-publish-build.sh` script is now more generic `deploy-build.sh` and supports both platforms
- `ARTIFACT_ID` was changed to `powerauth-android-sdk`, so the full gradle dependency definition will looks like: `compile 'io.getlime.security.powerauth:powerauth-android-sdk:0.15.0' ` 
- I removed following settings from `gradle.properties`file: 
  - `GROUP` - looks like this is not used in the project
  - `VERSION_CODE` - **android:versionCode** is not required for libraries, so we're setting this value to 1


proj-android/PowerAuthLibrary/build.gradle
```diff
@@ -75,5 +79,21 @@ publishing {
...
 
-
-
+bintray {
+    user = System.getenv('BINTRAY_USER')
+    key  = System.getenv('BINTRAY_API_KEY')
+    publications = ['maven']
+	pkg {
+		repo = 'PowerAuth'
+		name = 'powerauth-android-sdk'
```
ForestTree: Use project.ARTIFACT_ID here instead of this direct const string declaration.

proj-android/PowerAuthLibrary/build.gradle
```diff
@@ -75,5 +79,21 @@ publishing {
...
+		userOrg = 'lime-company'
```
ForestTree: I think that new item like "ORG" from 'gradle.properties' used here will be nice ;)

proj-android/PowerAuthLibrary/build.gradle
hvge: Thanks for hint. I'll use a separate properties for these values. In fact, that "name" has slightly different meaning than `ARTIFACT_ID`. It's just a package name where the archive will be uploaded.

PR #54, opened 2017-06-15
Android: Issue #36: Receive custom attributes from server

### LINKED ISSUE #36: Not possible to received customAttributes from response
We call createActivation with custom attributes and we need to receive customAttributes from server (some text to show on screen, etc.). Unfortunately, the ActivationCreateCustomResponse isn't delivered to any listener which we can use. It would be great when there is new ICreateCustomActivationListener with onCustomActivationCreateSucced(ActivationCreateCustomResponse response) and it's passed as argument to createActivation. This issue is in java client library as well (guys use it for integration tests on server).

syru88:
Fixes #36


hvge:
I have reimplemented this feature in PR #61

PR #71, opened 2017-11-24
Fix #70: iOS: createActivation error userInfo propagation

petrdvorak:
@hvge We can merge this change as a hotfix - it is pretty minor. I believe the proper fix would be using the more high-level `postToUrl:...` method of `PA2Client` that does similar work when serializing request/response objects.

PR #72, opened 2017-11-24
IOS: Fixed #70: Hotfix for additional error info propagation when custom activation fails on the server

### LINKED ISSUE #70: iOS: createActivation error userInfo propagation
When PowerAuthSDK.sharedInstance().createActivation returns error in callback, such NSError does not contain any data in userInfo dictionary, even when there should be some.

hvge:
This is similar change to #71 but with more safe data handling. In my implementation, the response dictionary is not set directly to the error's userInfo, but under the dedicated key (`PA2ErrorInfoKey_AdditionalInfo`). That's more traditional approach on IOS when error needs to keep an additional data.

I have also fixed other possible JSON deserializer crashes on nil data.


radekfrolik:
Tested against pod release PowerAuth2 0.17.0, working correctly, thanks

PR #74, opened 2017-11-27
Feature/token based auth (IOS)

hvge:
This pull request adds a support for token based authentication to IOS platform. It's a relative big change but it also contains a following important changes:

- Added a new `ECIESEncryptor` which is a E2E encryption base for our future protocol improvements
- Improved JSON response parsing in our networking. Now it's not possible to receive a wrong typed object in response objects.
- Many small improvements in private parts of the SDK


proj-xcode/Classes/keychain/PA2Keychain.m
```diff
@@ -291,12 +290,7 @@ - (PA2KeychainStoreItemResult) implAddValue:(NSData*)data forKey:(NSString*)key
 	[query setValue:@YES	forKey:(__bridge id)kSecUseNoAuthenticationUI];
 	
 	// If the system version is iOS 9.0+, use Touch ID if requested (kSecAccessControlTouchIDAny), or use kNilOptions
-	SecAccessControlCreateFlags flags = kNilOptions;
-	if ([PA2Keychain canUseTouchId]) {
-		if (useTouchId) {
-			flags = kSecAccessControlTouchIDAny;
-		}
-	}
+	SecAccessControlCreateFlags flags = useTouchId ? kSecAccessControlTouchIDAny : kNilOptions;
```
petrdvorak: We removed call to `canUseTouchId` here - is this correct? This was there to prevent using the flag on devices without fingerprint scanner. The `useTouchId` flag represents programmers desire to use Touch ID, `canUseTouchId` checks if this is possible...

proj-xcode/Classes/networking/entity/PA2EncryptedRequest.h
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#import "PA2NetworkObject.h"
+
+@class PA2ECIESEncryptor;
```
petrdvorak: Why do we need this class statement here? Seems unnecessary in the model class...

proj-xcode/Classes/keychain/PA2Keychain.m
hvge: That's correct, because it's already evaluated several lines above.

proj-xcode/Classes/networking/entity/PA2EncryptedRequest.h
hvge: Right, that's redundant. Previously, there was a reference to the `PA2ECIESCryptogram` & that enryptor, but I've changed that code. I'll fix this.

proj-xcode/Classes/keychain/PA2Keychain.m
hvge: It's funny, because you would like to save several object allocations (as comment at the beginning of the method is saying), but in fact, that redundant call did allocate several unnecessary objects... :)

petrdvorak: OK - Sorry, I see that now, after I opened the whole file. The old impl was some leftover check...

PR #77, opened 2017-12-04
Feature/token based auth android

hvge:
This pull requests adds support for token-based authentication to Android platform. On top of that, there are small fixes and updates:
-  I've fixed some compatibility in `Otp` class (which I broke by accident in previous commits, but we fortunately didn't release a library from that code). 
- The `PA2Keychain` has a new interface for getting & storing strings.
- Simplified request creation in `PA2Client`

### DIFF
--- a/proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthSDK.java
+++ b/proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthSDK.java
@@ -100,6 +100,7 @@ public class PowerAuthSDK {
     private PA2EncryptorFactory mEncryptorFactory;
     private PA2Keychain mStatusKeychain;
     private PA2Keychain mBiometryKeychain;
+    private PowerAuthTokenStore mTokenStore;
 
     /**
      * Helper class for building new instances.
@@ -145,7 +146,7 @@ public class PowerAuthSDK {
             } else {
                 instance.mClientConfiguration = new PowerAuthClientConfiguration.Builder().build();
             }
-            instance.mClient = new PA2Client();
+            instance.mClient = new PA2Client(mConfiguration, mClientConfiguration);
 
             instance.mStatusKeychain = new PA2Keychain(instance.mKeychainConfiguration.getKeychainStatusId());
             instance.mBiometryKeychain = new PA2Keychain(instance.mKeychainConfiguration.getKeychainBiometryId());
@@ -295,7 +296,7 @@ public class PowerAuthSDK {
             headers.put(httpHeader.getKey(), httpHeader.getValue());
 
             // Perform the server request
-            return mClient.vaultUnlockSignatureHeader(mConfiguration, mClientConfiguration, headers, new INetworkResponseListener<VaultUnlockResponse>() {
+            return mClient.vaultUnlock(headers, new INetworkResponseListener<VaultUnlockResponse>() {
 
                 @Override
                 public void onNetworkResponse(VaultUnlockResponse vaultUnlockResponse) {
@@ -313,6 +314,23 @@ public class PowerAuthSDK {
             listener.onFetchEncryptedVaultUnlockKeyFailed(new PowerAuthErrorException(httpHeader.getPowerAuthErrorCode()));
             return null;
         }
+    }
+
+    /**
+     * Returns reference to {@code PowerAuthTokenStore} instance. The internal instance is created on demand, when
+     * the getter is called for first time.
+     *
+     * @return Reference to {@code PowerAuthTokenStore} instance.
+     */
+    public PowerAuthTokenStore getTokenStore() {
+        if (mTokenStore == null) {
+            synchronized (this) {
+                if (mTokenStore == null) {
+                    mTokenStore = new PowerAuthTokenStore(this, mStatusKeychain, mClient);
+                }
+            }
+        }
+        return mTokenStore;
     }
 
     /**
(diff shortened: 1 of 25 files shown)


proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthSDK.java
```diff
@@ -315,6 +316,23 @@ public void onNetworkError(Throwable t) {
         }
     }
 
+    /**
+     * Returns reference to {@code PowerAuthTokenStore} instance. The internal instance is created on demand, when
+     * the getter is called for first time.
+     *
+     * @return Reference to {@code PowerAuthTokenStore} instance.
+     */
+    public PowerAuthTokenStore getTokenStore() {
```
petrdvorak: Not sure if this is correct, see https://en.wikipedia.org/wiki/Double-checked_locking#Usage_in_Java ("// Broken multithreaded version")...

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthTokenStore.java
```diff
@@ -0,0 +1,421 @@
...
+        final PowerAuthAuthorizationHttpHeader authHeader = sdk.requestSignatureWithAuthentication(context, authentication, "POST", PA2CreateTokenEndpoint.CREATE_TOKEN, postData.first);
+        if (!authHeader.isValid()) {
+            listener.onGetTokenFailed(new PowerAuthErrorException(authHeader.powerAuthErrorCode));
+            return null;
+        }
+        final HashMap<String, String> headers = new HashMap<>();
+        headers.put(authHeader.getKey(), authHeader.getValue());
+
+        // 4) Execute HTTP request
+        return httpClient.createToken(headers, request, new INetworkResponseListener<TokenCreateResponse>() {
```
petrdvorak: Note: You were right our networking could be greatly improved here. The signature header computation basically relies on the same serialization routine (object -> bytes) working the same way twice...

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthTokenStore.java
```diff
@@ -0,0 +1,421 @@
...
+     * @param authentication An authentication instance specifying what factors should be used for token creation.
+     * @param listener Listener with callbacks to receive a token.
+     * @return {@code AsyncTask} associated with the running server request or null if request has been processed synchronously.
+     */
+    public @Nullable AsyncTask requestAccessToken(@NonNull final Context context, @NonNull final String tokenName, @NonNull PowerAuthAuthentication authentication, @NonNull final IGetTokenListener listener) {
+
+        Throwable error = null;
+        PowerAuthPrivateTokenData tokenData = null;
+
+        synchronized (this) {
```
petrdvorak: Note: I admit it is probably my lack of knowledge here, but why do we need to synchronize this code here? Result of call to "canRequestForAccessToken" may be invalidated from outside this method, call to "getTokenData" may fail despite this synchronization. Or are we trying to synchronize only call to "getTokenData" here?

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthTokenStore.java
```diff
@@ -0,0 +1,421 @@
...
+     * @param tokenName Name of token to be removed
+     * @param listener Listener with callbacks.
+     * @return {@code AsyncTask} associated with the running server request or null in case of error.
+     */
+    public @Nullable AsyncTask removeAccessToken(@NonNull final Context context, @NonNull final String tokenName, @NonNull final IRemoveTokenListener listener) {
+
+        Throwable error = null;
+        PowerAuthPrivateTokenData tokenData;
+
+        synchronized (this) {
```
petrdvorak: Note: Similar question as above - it seems we only try to synchronize "getTokenData" - can we handle it in the method itself?

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthSDK.java
hvge: My mistake, I've fixed that and the method is fully synchronized.

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthTokenStore.java
hvge: You're correct, but if you invalidate the session during the getting token from the server, then the created token still cannot be used afterwards. That's OK and predictable behaviour. So, I need that lock only during accessing the private data structures (e.g. hash map). Despite of that, your hint had a good point and as a result, I've improved the code a little bit. Now the token is not stored to the keychain when is received from the server and the SDK has no longer a valid activation.

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthTokenStore.java
hvge: It would also work if I declare the method as synchronized, but it may be confusing for the API user, that the consistency is somehow guaranteed also during the internal network operation. That's not true and in fact, it's difficult to achieve (because we don't want to lock the whole object during the network task execution).

The whole `PowerAuthTokenStore` class is thread safe (e.g. you can't damage its internal data by accessing from multiple threads) but it's not aware agains the simultaneous asynchronous operations executed for the same token name. This is mentioned in the documentation.

petrdvorak:
(Please merge whenever you feel comfortable.)

PR #80, opened 2018-01-09
Feature/ios extensions

hvge:
This pull request adds a new library providing "baby" PowerAuth for IOS extensions. The new library is distributed as `PowerAuth2ForExtensions` cocoa pod, which provides a precompiled static, FAT framework.

Note that the change also contains files for watchOS library, but it's not finished yet. Support for watchOS will be added in separate PR.


PowerAuth2ForExtensions.podspec
```diff
@@ -0,0 +1,33 @@
+Pod::Spec.new do |s|
+	# General information
+	s.name              = 'PowerAuth2ForExtensions'
+	s.version           = '0.1.3'
```
hvge: don't worry about the version, it will be regenerated when the production pod will be published.

PR #83, opened 2018-01-26
watchOS SDK bugfixes & Improvements

hvge:
This pull request addresses several bugs which I recently found in watchOS and its counterpart IOS SDK:

- Fixed packet handling when watchOS requests IOS for token and the token doesn't exist.
- Improved error reporting when watchOS requests for token and IOS counterpart has no longer a valid activation.
- PowerAuthTokenStore on watchOS now can requst for tokens without using PowerAuthAuthentication
- If activation is removed from watchOS, then the associated tokens are also removed from keychain on watchOS
- Various small improvements, like more safe cancellable object impl., better logging, etc...


petrdvorak:
Feel free to merge.

PR #89, opened 2018-02-13
IOS: Fix #88: PowerAuthSDK doesn't depend on shared configs.

### LINKED ISSUE #88: Break PowerAuthSDK dependency on shared configurations
Currently, our iOS `PowerAuthSDK` class internally depends on several shared configuration instances: `PA2ClientConfiguration.sharedInstance()` & `PA2KeychainConfiguration.sharedInstance()`. We should break this dependency and allow object instantiation with using custom configuration objects.

### Benefits

* I'm working on [LimeAuth](https://github.com/lime-company/swift-lime-auth) library where I would like to allow instantiate high level object with all application provided configurations. Right now, it's kind of confusing when developer has to provide own `PowerAuthConfiguration` and then access shared instances for networking, etc... 
* Better thread safety. Later changes to shared configurations doesn't affect already instantiated SDK objects.
* Better object testability

### Proposed changes

* Both, `PA2ClientConfiguration` and `PA2KeychainConfiguration` should implement `NSCopying` protocol. 
* Add new read-only `keychainConfiguration` property to `PowerAuthSDK`. The property should return copy of internally used configuration, to do not allow post-inistantiation changes in keychain configuration. 
* Add new read-only `httpClientConfiguration` property to `PowerAuthSDK`. The property should return copy of internally used configuration, to do not allow post-inistantiation changes in client configuration. 
* Add new designated initializer to `PowerAuthSDK` to allow construct object with exact keychain & client configurations. Due to backward compatibility, we should provide fallbacks to shared config instances once one of the new parameters is nil
* Fix `PA2PrivateHttpTokenProvider` which also depends on `PA2ClientConfiguration.sharedInstance()`

hvge:
See issue #88 for more details

### DIFF
--- a/proj-xcode/Classes/keychain/PA2KeychainConfiguration.h
+++ b/proj-xcode/Classes/keychain/PA2KeychainConfiguration.h
@@ -43,7 +43,7 @@ extern NSString * __nonnull const PA2KeychainKey_Possession;
 
 /** Class that is used to provide default (shared) Keychain storage configuration.
  */
-@interface PA2KeychainConfiguration : NSObject
+@interface PA2KeychainConfiguration : NSObject<NSCopying>
 
 /** Access group name used by the PowerAuthSDK keychain instances.
  */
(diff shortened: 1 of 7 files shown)


petrdvorak:
We should change documentation a bit - currently, the code uses shared instances that can change over the time and change is reflected. This is no longer the case since we use copies.

hvge:
You're right. I'll update documentation and mention this fact in release notes for `0.18.0`

Issue #91, opened 2018-02-22
IOS: Strip simulator architecture on Archive task


hvge:
This issue is related to #81 

We recently changed how our iOS library is distributed, from static FAT library, to a static framework. Meanwhile, I had to change that for once more to a dynamic framework, due to fact, that static fw cannot be linked to another dynamic fw. This kind of transitional dependency is forbidden (by cocoapods, but it's rational, you don't want to have linked static library multiple times in multiple dynamic frameworks)

So, once I changed `PowerAuth2.framework` to a dynamic, I've soon realized that our customers will not be able to publish their applications to iTunes Store, due to forbidden architectures still present in precompiled framework (check for example [this question on SO](https://stackoverflow.com/questions/30547283/submit-to-app-store-issues-unsupported-architecture-x86)). Carthage guys had [the same problem](https://github.com/Carthage/Carthage/issues/188).

I personally feel, that switch to the dynamic framework is a good thing for PowerAuth. It allows us to develop a libraries on top of PowerAuth while application's code can still use all PowerAuth features. This is for example the only one way how make [LimeAuth library](https://github.com/lime-company/swift-lime-auth) possible. 

So, we have several options how to solve this problem.

### Carthage way

We can adopt for example a Carthage way, when the developer must add some pre-archive script into the application's project, which removes x86 architectures from all dynamic frameworks.

This is quick solution.
 
### Split library

We can split PowerAuth mobile SDK into two separate projects:
- Low level core, (c++, openssl + thin objc wrapper) which will be distributed as static FAT library
- High level code, which will be distributed in pod as source code

This is a good solution, because cocoapods will manage that `PowerAuth2.framework` will be fully debuggable (#76) and will be compiled only for an actually required architectures (e.g. Archive task will compile just ARM code, build for simulator will use x86, etc..). Linking static fat library against another target is correct, because the target dictates which architectures has to be added to the final executable (or library).

Unfortunately, I'm not sure if this is doable right now (see my [comment in issue 76](https://github.com/lime-company/powerauth-mobile-sdk/issues/76#issuecomment-355061114))

### Extensions & watchOS

We don't have this problem on app extensions or watchOS. That frameworks are static and therefore can be linked only to extensions' target.

hvge:
Fixed by PR #96

Issue #76, opened 2017-11-29
Allow easier debugging of iOS SDK


petrdvorak:
Debugging SDK internals in case something does not work is incredibly painful. We need to improve this. Good way could be to include source files with `PowerAuthSDK-Debug` CocoaPod, so that integrators can place a breakpoint in it and tell us more specifically where the issue is...

hvge:
This is IMHO redundant to #49 From my own experience as an user of PA-SDK, the errors reported to the application is the biggest pain in the ass. If you need to debug the library, then the library is basically wrong by design. Which is not in my opinion, but we need to improve SDK behaviour when something's going off the rails.

hvge:
Anyway, if we will do this, then we have to move C++ & "core" parts of SDK to the separate project file (e.g. separate static library), otherwise it will be hell on earth to force cocoapods to generate right projects with all dependencies on C++, cc7, openssl, etc... 

 It's more work that you would expect :)

hvge:
The enhancement is not doable, unless the cocoapods fix its crappy dependency management. So, I'm adding a "sleeping" tag and will take a look on this when cocoapods 1.5 will be shipped.

### tl;dr

The blocker is our own support for IOS Extensions (currently in development) which is in conflict with regular PowerAuth library for IOS. We're sharing headers and some source codes in libraries for IOS, watchOS and IOS extensions and that's exactly what pod tool is not capable to handle.

### Long explanation

The pod tool re-generates project file for each `pod update` run. The main problem is that generated `Pods.xcodeproj` doesn't respect dependencies between pods and sets public pod's headers visible for each other pod used in the application (or application's target). This behaviour has a significant impact when both `PowerAuth2` and `PowerAuth2ForExtensions` pods are used in the application. These two libraries contains several shared source codes and importantly, are exposing several public headers with the same name. Due to fact that each pod can access headers from each other, the compiler then fails on duplicate interface definitions.

This happens independently on what pods are used in application's targets and leads actually to two different problems:

1. the problem with compiling pods as I described above
2. the problem with final application linking, when pod tool currently sets dependency on our both libraries to all available targets. That's really stupid and will be probably fixed in cocoapods 1.5

So, looks like that it will be easier to workaround current problems and later rewrite the high level codes to swift, which has a better module management. So that's why this is sleeping now...

petrdvorak:
OK, makes sense.

hvge:
Fixed by PR #96 

[...]

PR #96, opened 2018-04-05
Issue/91 split ios library

hvge:
This change brings a complete rework how `PowerAuth2.framework` is compiled:
- Now there are two separate projects: 
   - `PowerAuthCore.xcodeproj` containing low level C++ code & thin ObjC wrapper on top of that.
   - `PowerAuthLib.xcodeproj` containing high level SDK source codes. The project depends on `PowerAuthCore.xcodeproj` (it's already embedded in the project tree)
- Script for CocoaPods build preparation now produces:
  - Static FAT library `libPowerAuthCore.a` with all architectures
  - Prepares all public & private source codes into build output folder, so our podspec can easily capture them
- `pod PowerAuth2-Debug` is now deprecated

The change is resolving following issues:
- #91 - we don't use precompiled framework anymore, so we don't have a problem with x86 architectures in iTunesConnect
- #76 - the main part of SDK is now distributed as source codes, so it's more debug friendly


scripts/ios-build-libraries.sh
```diff
@@ -8,15 +8,27 @@ set +v
...
-# one file.
-# 
+# The result of the build process is:
+#    libPowerAuthCore.a:
+#      multi-architecture static library (also called as "fat") with all 
+#      core functionality of PowerAuth2 SDK. The library contains all C++
+#      code, plus thin ObjC wrapper written on top of that codes.
+#
+#    SDK sources:
+#      all SDK higl level source codes are copied to destination directory.
```
petrdvorak: Typo in comment - "high" vs. "higl"

hvge: I removed that files just now. That was crap from xcode...

hvge: Fixed that typo.

PR #127, opened 2018-07-03
IOS: Provide more info about biometry and its status on the system

### LINKED ISSUE #125: IOS: Provide more info about biometry and its status on the system
Currently, we have a several ways for determining whether the biometry is available on the system, or is enabled for PA activation:

- At the system level, there are two static properties in `PA2Keychain`:
   - `PA2Keychain.canUseBiometricAuthentication` - returning YES whether biometry can be used right now
  - `PA2Keychain.supportedBiometricAuthentication` - returning supported type of biometric authentication, or "none", if there's no such support, or biometry cannot be used right now.

- At activation level:
  - `PowerAuthSDK.hasBiometryFactor()` - returning YES if signature can be calculated with biometric factor. 

This is in general OK for most of the scenarios, but if the application wants to handle for example a biometry lockdown, or other corner states, then it has to implement its own state detection. In my own experience, if you want to handle all states properly for also older IOS versions, then it leads to a very dirty swift code (like forcing off deprecation warnings, due to deprecated constants in `LocalAuthentication` fw). 

My proposal is to enhance the existing `PA2Keychain` interface, with getters providing an additional information about biometry state. We should cover a following new states:

- biometry authentication is available, but is locked due to too many failed attempts
- biometry authentication is available, but has not been enrolled yet

We should keep the behavior of current interfaces intact, because those are in general very handy for most of the application scenarios.

hvge:
This PR brings following changes:

- `PA2SupportedBiometricAuthentication` is now deprecated and the replacement is `PA2BiometricAuthenticationType`
- added `PA2BiometricAuthenticationStatus` covering all possible states of biometry authentication
- added `struct PA2BiometricAuthenticationInfo` wrapping both "status" and "type" enumerations.
- added `PA2Keychain.biometricAuthenticationInfo` class property

Behavior of `PA2Keychain.canUseBiometricAuthentication` and `PA2Keychain. supportedBiometricAuthentication` should be the same as before.

Fixes #125

### DIFF
--- a/proj-xcode/Classes/keychain/PA2Keychain.m
+++ b/proj-xcode/Classes/keychain/PA2Keychain.m
@@ -265,11 +265,22 @@ static void _AddUseNoAuthenticationUI(NSMutableDictionary * query)
 #pragma mark - Biometry support
 
 /**
- Looks like Apple introduced a new biometry type. We should try to continue,
- and pretend that TouchID is available. Application's UI will probably display
- wrong information, but at least it may work.
+ Private helper function to convert LABiometryType enum into our PA2BiometricAuthenticationType
  */
-#define __BiometricAuthenticationType_Fallback 	PA2BiometricAuthenticationType_TouchID
+API_AVAILABLE(ios(11.0))
+static PA2BiometricAuthenticationType _LABiometryTypeToPAType(LABiometryType bt)
+{
+	if (bt == LABiometryTypeTouchID) {
+		return PA2BiometricAuthenticationType_TouchID;
+	} else if (bt == LABiometryTypeFaceID) {
+		return PA2BiometricAuthenticationType_FaceID;
+	}
+	// Looks like Apple introduced a new biometry type. We should try to continue,
+	// and pretend that TouchID is available. Application's UI will probably display
+	// wrong information, but at least it may work.
+	PA2Log(@"Warning: LAContext.biometryType contains unknown biometryType %@.", @(bt));
+	return PA2BiometricAuthenticationType_TouchID;
+}
 
 /**
  A private function returns full information about biometric support on the system. The method internally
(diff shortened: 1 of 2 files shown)


proj-xcode/Classes/keychain/PA2Keychain.m
```diff
@@ -264,33 +264,112 @@ - (NSDictionary*) allItemsWithPrompt:(NSString*)prompt withStatus: (OSStatus *)s
...
+			// Now check the type of biometry
 			if (@available(iOS 11.0, *)) {
 				LABiometryType bt = context.biometryType;
 				if (bt == LABiometryTypeTouchID) {
-					return PA2SupportedBiometricAuthentication_TouchID;
+					info.biometryType = PA2BiometricAuthenticationType_TouchID;
 				} else if (bt == LABiometryTypeFaceID) {
-					return PA2SupportedBiometricAuthentication_FaceID;
+					info.biometryType = PA2BiometricAuthenticationType_FaceID;
+				} else {
```
petrdvorak: We should probably have one more `else if` branch here, for `LABiometryTypeNone` value (see the [docs](https://developer.apple.com/documentation/localauthentication/labiometrytype?language=objc)). Currently, we may fallback to unknown biometry type even for this value (if, for some reason, `canEvaluate` is true at the same time)...

proj-xcode/Classes/keychain/PA2Keychain.m
```diff
@@ -264,33 +264,112 @@ - (NSDictionary*) allItemsWithPrompt:(NSString*)prompt withStatus: (OSStatus *)s
...
+				info.biometryType = PA2BiometricAuthenticationType_TouchID;
+			}
+			//
+		} else {
+			// In case of error we cannot evaluate, but the type of biometry can be determined.
+			NSInteger code = [error.domain isEqualToString:LAErrorDomain] ? error.code : 0;
+			if (@available(iOS 11.0, *)) {
+				// On iOS 11 its quite simple, we have type property available and status can be determied
+				// from the error.
+				LABiometryType bt = context.biometryType;
```
petrdvorak: This seems to be the same code as the one in case of no error. Could we simplify the code (use `context.biometryType` for type of biometry and `canEvaluate`/error for status)?

proj-xcode/Classes/keychain/PA2Keychain.m
hvge: That case will never happen, because biometry is already evaluated and "none" type makes no sense here.

proj-xcode/Classes/keychain/PA2Keychain.m
hvge: I was thinking about that, but in my opinion, it will lead to more complex and less readable code. That @availability check has its internal cost, so I was trying to reduce its usage to minimum. Current impl. requires only one check for IOS9 and another for IOS11 for both, error & success branch. 

But I can move that enumeration translation to another function, which will be probably inlined.

petrdvorak: OK, makes sense - let's leave the code as it is, the optimization (if any) would be marginal...

hvge: My last commit moved that repeated code block to translation function. It's much better I think.

PR #143, opened 2018-09-12
Fix #100: Update project setup to target Android Oreo

### LINKED ISSUE #100: Android: Update project setup
Versions of various build tools are rather outdated, there are newer setups being used now.
The things to update are: android build plugin, gradle wrapper, build repositories.

TomasKypta:
It was necessary to add new support lib dependency to fix
build issue made with the newer libs and targets.
Updated dependencies to the latest compatible stable versions.


petrdvorak:
Looks OK

Issue #147, opened 2018-09-14
iOS: return NSHTTPURLResponse in http requests


kober32:
`postToUrl` in `PA2Client` is returning `NSURLResponse` in completion block. It would be much more convenient to return `NSHTTPURLResponse` right away since the response is this type every time anyway.

hvge:
I don't want to fix this. We're going to change this interface in Crypto 3.0 related works, so this change will just complicate things during the future merge. On top of that, `PA2Client` should not be used in applications for networking. Now, it's public interface, but I would rather make it private in the future SDK versions, than make it more usable from swift. 

But I'll consider this proposition in that crypto 3.0 major update.

PR #148, opened 2018-09-14
IOS: returning NSHTTPURLResponse in PA2Client

kober32:
Solves issue #147


hvge:
See my comment added to the issue.

PR #155, opened 2018-10-03
Crypto3: IOS: Added a brand new networking code for internal SDK purposes

hvge:
Added:
- PA2HttpClient for high level networking
- PA2HttpRequest for request & response processing
- PA2RestApiEndpoint, defining one particular endpoint on the server
- PA2AsyncOperation, implementing asynchronous NSOperation
- PA2Encodable & PA2Decodable protocols as replacement for PA2NetworkObject

There's still old PA2Client code as a reference impl, but will be removed soon.


proj-xcode/Classes/networking/private/client/PA2AsyncOperation.m
```diff
@@ -0,0 +1,174 @@
...
+		_reportQueue = queue;
+		_lock = [[NSLock alloc] init];
+	}
+	return self;
+}
+
+
+#pragma mark - main & cancel
+
+- (void) main
```
kober32: Primary function to override should be `start`. Overriding only `main` will result in default `NSOperation.start` to be called, which is performing some internal state changes (that are overridden and thus don't reflect actual state) like `willChangeValueForKey:@"isExecuting"`

proj-xcode/Classes/networking/private/client/PA2HttpClient.m
```diff
@@ -0,0 +1,131 @@
...
+		if ([baseUrl hasSuffix:@"/"]) {
+			_baseUrl = [baseUrl substringToIndex:baseUrl.length - 1];
+		} else {
+			_baseUrl = baseUrl;
+		}
+		
+		// Prepare queues.
+		// Note that dispatch queue is serial, but NSOPerationQueue is concurrent (by default)
+		_dispatchQueue = dispatch_queue_create("PA2HttpClient", DISPATCH_QUEUE_SERIAL);
+		_operationQueue = [[NSOperationQueue alloc] init];
```
kober32: You can use `_operationQueue.maxConcurrentOperationCount = 1;` to create serial `NSOperationQueue`

proj-xcode/Classes/networking/private/client/PA2AsyncOperation.m
```diff
@@ -0,0 +1,174 @@
...
+	[self cancelImpl];
+	//
+	[_lock unlock];
+}
+
+- (void) cancelImpl
+{
+	if (!_inExecutionBlock) {
+		if (_cancelBlock) {
+			_cancelBlock(self, _operationTask);
```
kober32: I would expect the `cancelBlock` to be called on `reportQueue` too. Both cancel and complete are basically "report result" functions.

proj-xcode/Classes/networking/private/client/PA2HttpClient.m
hvge: That's intended. I don't want to bring serialization queue now.

PR #168, opened 2018-12-05
Feature/130 request interceptors

### LINKED ISSUE #130: Add ability to add the custom HTTP headers to standard RESTful API calls
Sometimes, the server that publishes the PA Standard RESTful API may run behind an API Management Gateway / Proxy that requires an authentication of various sorts (like an API-key, or basic HTTP, ...).

pa-proxy

Currently, there is no way to call the standard server endpoint with custom HTTP headers from the mobile SDK.

We should add a mechanism to cover all use-cases that call PA Standard RESTful API so that we allow customizing HTTP headers of the request. We should also make sure that by doing this, the SDK user does not override our HTTP headers, mainly the `X-PowerAuth-Authorization` header or `Content-Type` header.

Let's analyze the best way to approach this issue:

- adding the custom headers to every call?
- adding the authentication strategy to the HTTP client (as we do with SSL handling)?
- other? :)

hvge:
This PR brings a new interfaces to our HTTP client configuration, allowing modify the internally created request, before its execution. Check attached issue #130 for details.

Close #130


proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/networking/interceptors/BasicHttpAuthenticationRequestInterceptor.java
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * Builds a Basic HTTP Authentication header's value for given username and password.
+     *
+     * @param username String with username
+     * @param password String with password
+     * @return Value for Basic HTTP Authentication header
+     */
+    private static @NonNull String buildHeaderValue(@NonNull String username, @NonNull String password) {
+        final byte[] payload = (username + ":" + password).getBytes(Charset.defaultCharset());
+        return "Basic " + Base64.encodeToString(payload, Base64.NO_PADDING);
```
TomasKypta: There should be Base64.NO_WRAP instead.

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/networking/interceptors/HttpRequestInterceptor.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+public interface HttpRequestInterceptor {
+
+    /**
+     * Called before the provided {@link HttpURLConnection} is executed. The implementation must
+     * count with that method is called from other than UI thread.
+     *
+     * @param connection {@link HttpURLConnection} object to be modified.
+     */
+    void processRequestConnection(@NonNull HttpURLConnection connection);
```
TomasKypta: So do we really intend to allow any kind manipulation with `HttpUrlConnection` (thus manipulation with `HttpsUrlConnection` too)?
Or do we want to allow only header addition?

This way everything on `Http(s)UrlConnection` can be touched and changed such as `SSLSocketFactory`, `HostnameVerifier`, request properties (timeouts, method, …), …
We are entering dangerous waters with this, devs can easily break things. Is it ok for us?

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/networking/interceptors/BasicHttpAuthenticationRequestInterceptor.java
hvge: You're right!

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/networking/interceptors/HttpRequestInterceptor.java
hvge: Well, my original idea was to add just a dictionary with an additional headers to the configuration, but @petrdvorak wanted this more "enterprise" :) It is the same on iOS, so yes, if our clients wants to shoot themselves in the foot, then lets allow that.

hvge: To be clear. This approach allows us to add a more advanced tweaks in the future, but we should cover this in the documentation, that creating custom interceptors and tweaking the connection is not recommended.

hvge: I've added warnings to inline documentation.

PR #173, opened 2018-12-06
Issues/170 public ecies

hvge:
This PR adds a public interface for our new ECIES based E2EE.

### DIFF
--- a/proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthSDK.java
+++ b/proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthSDK.java
@@ -1604,24 +1604,21 @@ public class PowerAuthSDK {
     /**
      * Creates a new instance of ECIES encryptor suited for application's general end-to-end encryption purposes.
      * The returned encryptor is cryptographically bounded to the PowerAuth configuration, so it can be used
-     * with or without the valid activation.
+     * with or without a valid activation.
      *
-     * @return New instance of {@link EciesEncryptor} object or {@code null} if {@link PowerAuthConfiguration}
-     *         contains an invalid data.
+     * @return New instance of {@link EciesEncryptor} object.
+     * @throws PowerAuthErrorException if {@link PowerAuthConfiguration} contains an invalid configuration.
+     *         You can call {@link PowerAuthErrorException#getPowerAuthErrorCode()} to get a more
+     *         detailed information about the failure.
      */
-    public @Nullable EciesEncryptor getEciesEncryptorForApplicationScope() {
-        try {
-            final IPrivateCryptoHelper helper = getCryptoHelper(null);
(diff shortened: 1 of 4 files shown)


proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthSDK.java
```diff
@@ -1599,4 +1599,46 @@ public void onFingerprintInfoDialogClosed() {
         dialog.show();
     }
 
+    // E2EE
+
+    /**
+     * Creates a new instance of ECIES encryptor suited for application's general end-to-end encryption purposes.
+     * The returned encryptor is cryptographically bounded to the PowerAuth configuration, so it can be used
+     * with or without the valid activation.
```
TomasKypta: with or without a valid activation

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthSDK.java
```diff
@@ -1599,4 +1599,46 @@ public void onFingerprintInfoDialogClosed() {
...
+     * @return New instance of {@link EciesEncryptor} object or {@code null} if {@link PowerAuthConfiguration}
+     *         contains an invalid data.
+     */
+    public @Nullable EciesEncryptor getEciesEncryptorForApplicationScope() {
+        try {
+            final IPrivateCryptoHelper helper = getCryptoHelper(null);
+            return helper.getEciesEncryptor(EciesEncryptorId.GENERIC_APPLICATION_SCOPE);
+        } catch (PowerAuthErrorException e) {
+            PA2Log.e("Failed to create application scoped ECIES encryptor: " + e.getMessage());
+            return null;
```
TomasKypta: Or maybe it should throw an InvalidStateException instead of returning null (in both ecies methods).

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthSDK.java
```diff
@@ -1599,4 +1599,46 @@ public void onFingerprintInfoDialogClosed() {
...
+        }
+    }
+
+    /**
+     * Creates a new instance of ECIES encryptor suited for application's general end-to-end encryption purposes.
+     * The returned encryptor is cryptographically bounded to the device's activation, so it can be used only
```
TomasKypta: …bounded to a device's activation, …

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthSDK.java
hvge: Right. I thought that we don't throw an exceptions from that object (because the class was rewritten from objc), but I was wrong. For this case it's OK to throw that `PowerAuthErrorException`

PR #196, opened 2019-01-29
Android: Fixed #114: Improved ICommitActivationWithFingerprintListener interface

### LINKED ISSUE #114: Fix inconsistent commitActivation* methods
There are four `commitActivation*` methods, three of them return int (error code) and the fourth swallows the error code.
Method `commitActivation` ignores internal error code coming from `commitActivationWithPassword`.
The callback for the method should be extended to contain error callback method.

hvge:
- Fixed missing `@PowerAuthErrorCodes` annotation

### DIFF
--- a/proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthSDK.java
+++ b/proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthSDK.java
@@ -733,6 +733,7 @@ public class PowerAuthSDK {
      * @throws PowerAuthMissingConfigException thrown in case configuration is not present.
      */
     @CheckResult
+    @PowerAuthErrorCodes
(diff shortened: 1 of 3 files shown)


proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthSDK.java
```diff
@@ -785,6 +789,7 @@ public void onFingerprintInfoDialogClosed() {
      */
     @RequiresApi(api = Build.VERSION_CODES.M)
     @CheckResult
+    @PowerAuthErrorCodes
```
TomasKypta: There's another `commitActivationWithPassword()` method, there too.

hvge: Right, fixed...

Issue #200, opened 2019-01-31
Android Crash in Fingerprint: "IV has already been used"


petrfuka:
Production crash in Fingerprint dialog.
Crash on old PowerAuth 0.16.5 but issue probably still exists.

Reported stacktrace:
```
Fatal Exception: java.lang.IllegalStateException: IV has already been used. Reusing IV in encryption mode violates security best practices.
       at android.security.keystore.AndroidKeyStoreUnauthenticatedAESCipherSpi.addAlgorithmSpecificParametersToBegin(AndroidKeyStoreUnauthenticatedAESCipherSpi.java:244)
       at android.security.keystore.AndroidKeyStoreCipherSpiBase.ensureKeystoreOperationInitialized(AndroidKeyStoreCipherSpiBase.java:237)
       at android.security.keystore.AndroidKeyStoreCipherSpiBase.engineDoFinal(AndroidKeyStoreCipherSpiBase.java:495)
       at javax.crypto.Cipher.doFinal(Cipher.java:1736)
       at io.getlime.security.powerauth.keychain.fingerprint.FingerprintAuthenticationHandler.encryptedKey(FingerprintAuthenticationHandler.java:236)
       at io.getlime.security.powerauth.keychain.fingerprint.FingerprintAuthenticationDialogFragment$5.run(FingerprintAuthenticationDialogFragment.java:331)
       at android.os.Handler.handleCallback(Handler.java:789)
       at android.os.Handler.dispatchMessage(Handler.java:98)
       at android.os.Looper.loop(Looper.java:164)
       at android.app.ActivityThread.main(ActivityThread.java:6944)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:327)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1374)
```

hvge:
Do you have more information about the device? This might be a device specific issue.

hvge:
[...]

PR #227, opened 2019-06-21
Issues/200 biomeric dialog crash

hvge:
Android: Hotfix #200: Crash due to possible "IV has already been used." exception.

This change also fixes wrong result handling in the FingerprintAuthenticationDialogFragment.

### DIFF
--- a/proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/keychain/fingerprint/FingerprintAuthenticationHandler.java
+++ b/proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/keychain/fingerprint/FingerprintAuthenticationHandler.java
@@ -46,7 +46,6 @@ public class FingerprintAuthenticationHandler extends FingerprintManager.Authent
     private final FingerprintManager mFingerprintManager;
     private CancellationSignal mCancellationSignal;
     private FingerprintManager.CryptoObject mCryptoObject;
-    private SecretKey mSecretKey;
     private FingerprintKeystore mKeyStore;
     private Cipher mCipher;
     private boolean mForceGenerateNewKey;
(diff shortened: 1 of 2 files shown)


proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/keychain/fingerprint/FingerprintAuthenticationHandler.java
```diff
@@ -246,21 +236,31 @@ public FingerprintStage initCrypto() {
...
+    /**
+     * Contains cached encrypted key. The value is calculated only for once.
+     */
+    private byte[] alreadyEncryptedKey;
+
+    /**
+     * Contains true if {@link #alreadyEncryptedKey} has been already constructed.
+     * The purpose of this flag is to prevent duplicate calls to {@code mCipher.doFinal()}.
+     */
+    private boolean hasAlreadyEncryptedKey;
```
petrdvorak: Do we need this property, or can we just work with `alreadyEncryptedKey`? As a syntactic sugar, we can have:

```java
private boolean hasAlreadyEncryptedKey() {
    return alreadyEncryptedKey != null;
}
```

hvge: I wanted to avoid double call to `doFinal()` on `Cipher` (original source of crash). It can be reproduced by double "success" reported by the `FingerprintManager`. I guess that some devices can report multiple "success" states quickly in relative short time. 

This boolean guarantees that it's not called for twice even if the first attempt failed and the result is `null`. Perhaps it's overcomplicated. 

This is completely reworked In upcoming `1.2.0`, so the object receiving events from `FingerprintManager` will never report this situation for twice.

petrdvorak: OK, understood...

PR #277, opened 2020-02-04
Android: Fix #276: Support for Android 10 biometry

### LINKED ISSUE #276: Add support for Android 10 biometry
Our current support for android biometry is based on `BiometricPrompt` API, introduced in Android 9. Unfortunately, Android 9 doesn't provide proper API to evaluate whether the biometry's enrolled on the system. Currently, we're using hack with using old `FingerprintManager` to do that. This approach doesn't work on Android 10, in case that non-fingerprint biometry is enrolled on the system. We have to use new [BiometricManager](https://developer.android.com/reference/android/hardware/biometrics/BiometricManager) class to proper enumeration on newer systems.

This enhancement will be implemented for SDK `1.3.x` and back-ported to `1.2.x`.

hvge:


proj-android/build.gradle
```diff
@@ -12,7 +12,7 @@ buildscript {
 }
 
 ext {
-    compileSdkVersion = 28
+    compileSdkVersion = 29
     targetSdkVersion = 28
```
TomasKypta: We should probably target the latest.

hvge: Right, I'll change that.

PR #293, opened 2020-04-27
Android: Fix #292: Workaround for biometry on Samsung devices

### LINKED ISSUE #292: Biometry workaround for Samsung
According to `androidx.biometric` [support library](https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/biometric/biometric/src/main/res/values/devices.xml), we should fallback `BiometricPrompt` on several Samsung devices to deprecated `FingerprintManager`. This is due to the fact, that such devices don't provide enough biometric security for one of the supported factors (typically Face detection). 

Related [SO discussion](https://stackoverflow.com/questions/55634812/biometricprompt-crashes-on-samsung-s9-with-face-unlock), and [DeviceConfig.java](https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/biometric/biometric/src/main/java/androidx/biometric/DeviceConfig.java) support class. 

This fix will be backported to `1.2.x` and `1.3.x` SDK branches.

hvge:


proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/biometry/impl/BiometricHelper.java
```diff
@@ -134,4 +136,78 @@
             return null;
         }
     }
+
+    /**
+     * Determine if the current device should explicitly fallback {@code FingerprintManager} based
+     * authentication.
```
petrdvorak, line 142: I would improve the comment here to explain what is the idea - it took me some time to build a Venn diagram in my head for the conditions below. Also, I would look if we could write the conditions easier (specifically, I was struggling with the negative comparison for Android P, and figuring out what the condition means).

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/biometry/impl/BiometricHelper.java
```diff
@@ -134,4 +136,78 @@
...
+        }
+        if (!isVendorInFallbackList(context)) {
+            return false;
+        }
+        return isModelInPrefixList(context.getResources().getStringArray(R.array.crypto_fingerprint_fallback_prefixes));
+    }
+
+    /**
+     * Determine if the current device should hide a fingerprint dialog immediately. This is required
+     * for the problematic devices that display it's own, custom fingerprint dialog overlay.
```
petrdvorak, line 159: Also, I would improve the comment a bit to explain what is going on inside, or simplified the conditions.

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/biometry/impl/BiometricHelper.java
kober32: I second to that, the comparison with the P looks out-of-place without explanation.

hvge: OK, I explained that in more detail in my last commit.

PR #321, opened 2020-08-03
Android integration tests

hvge:
This PR adds following changes to Android platform:

- Framework for integration tests
- Initial bunch of integration tests

### DIFF
--- a/proj-android/PowerAuthLibrary/build.gradle
+++ b/proj-android/PowerAuthLibrary/build.gradle
@@ -20,6 +20,7 @@ android {
         }
 
         testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
+        loadInstrumentationTestConfigProperties(project, owner)
     }
 
     compileOptions {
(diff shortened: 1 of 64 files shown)


kober32:
Hmm, 2 tests are failing for me (running against WUAT server)

```
io.getlime.security.powerauth.integration.tests.RecoveryActivationTest > testCreateActivationWithNoRC[Pixel_3_API_29(AVD) - 10] FAILED 
        java.lang.Exception: Recovery config for application 606 is not updated after successful response.
        at io.getlime.security.powerauth.integration.support.client.PowerAuthClientV3.updateRecoveryConfig(PowerAuthClientV3.java:140)

io.getlime.security.powerauth.integration.tests.RecoveryActivationTest > testCreateRecoveryActivation[Pixel_3_API_29(AVD) - 10] FAILED 
        java.lang.Exception: Recovery config for application 606 is not updated after successful response.
        at io.getlime.security.powerauth.integration.support.client.PowerAuthClientV3.updateRecoveryConfig(PowerAuthClientV3.java:140)
```

also, before I could run the tests in the first place, I had to provide an `AndroidManifest.xml` file inside the `androidTest` folder with enabled `HTTP`. Maybe this should be part of the repo? I guess most of the time, the test will be running against `HTTP` and not `HTTPS`.

```
<?xml version="1.0" encoding="utf-8"?>
<manifest xmlns:android="http://schemas.android.com/apk/res/android" package="io.getlime.security.powerauth">

    <application android:usesCleartextTraffic="true" />

</manifest>

```

PR #326, opened 2020-08-25
Android: Fix #325: Better Base64 string detection.

### LINKED ISSUE #325: EncryptedKeychain crash when migrating from legacy keychain
The `EncryptedKeychain` crash while migrating data from legacy, unencrypted keychain. The issue is caused by a wrong Base64 string detection in the import function:
```
Caused by: java.lang.IllegalArgumentException: bad base-64
at android.util.Base64.decode(Base64.java:162)
at android.util.Base64.decode(Base64.java:137)
at android.util.Base64.decode(Base64.java:119)
at io.getlime.security.powerauth.keychain.impl.EncryptedKeychain.importFromLegacyKeychain(EncryptedKeychain.java:325)
at io.getlime.security.powerauth.keychain.KeychainFactory.createKeychain(KeychainFactory.java:114)
at io.getlime.security.powerauth.keychain.KeychainFactory.getKeychain(KeychainFactory.java:67)
...
```
The problem is that `Base64.decode()` can throw an `IllegalArgumentException` in case that the provided string has bad padding. We should catch this exception and treat such string as non-Base64 data.

hvge:


petrdvorak (approved):
Looks OK

PR #330, opened 2020-08-31
Docs: Fix #329: Added information about activation states

### LINKED ISSUE #329: Document the need to remove activation locally
Even if the activation was removed on the server, the client-side developer also needs to remove it locally on the device. This causes some minor confusion among the developers. We should document this, probably in the chapter related to obtaining the status.

hvge:


petrdvorak (approved):
Looks good

PR #334, opened 2020-11-02
Android: Fix #332: Support for asymmetric biometric key

### LINKED ISSUE #332: Use asymmetric cipher for Android biometric factor protection
We can improve our biometric support on Android by using RSA cipher with key stored in the Android KeyStore. The idea behind this enhancement is that creating the asymmetric key-pair in the KeyStore and protecting the factor-related key, doesn't require immediate biometric authentication. This is due the fact, that we can encrypt data with using the public key immediately after the key-pair is created. That allows us to to turn-on biometry without actual use of biometric authentication. The behavior will be similar to iOS implementation.

So, this enhancement will bring the following changes:

- Add option to keychain, that enables RSA cipher for biometric factor protection. 
- By default, we'll still use old fashioned AES KDF that use biometric authentication also for enabling biometry process.

This enhancement will be implemented for `1.4.x` branch.

hvge:
This PR adds support for biometric factor protected by asymmetric RSA key-pair, stored in Android keystore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ In order to connect to the [PowerAuth](https://www.wultra.com/product/powerauth-
 
 If you need to upgrade PowerAuth Mobile SDK to a newer version, you can check following migration guides:
 
+- [Migration from version `1.4.x` to `1.5.x`](docs/Migration-from-1.4-to-1.5.md)
 - [Migration from version `1.3.x` to `1.4.x`](docs/Migration-from-1.3-to-1.4.md)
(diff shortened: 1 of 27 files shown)


kober32 (approved):
Looks and works OK.

Tested on already created activation with the previous SDK version and everything works, as well as new activation on the same device.

Tested on Samsung A5 with Android 8 and Fingerprint reader.

TomasKypta (approved):
Looks ok

PR #349, opened 2020-12-14
Android: Fix #348: Biometry lockout enhancements

### LINKED ISSUE #348: Propagate biometry lockout error back to application
The current biometric implementation on Android detects biometry lockout, but doesn't propagate such error back to the application. We should improve this, but at first, let me explain how it works now.

The "lockout" error is currently handled in two different ways, depending on when the lockout happened:

1. If the biometric sensor is locked out in the current authentication attempt (e.g. user tried multiple attempts, but failed)
   - In this particular situation, we report `PA2ErrorCodeBiometryNotRecognized` error code, but such code is never reported back to the application. The higher levels of SDK catch this error and generate an invalid PowerAuth signature, to raise the failed attempts counter on the server. 
2. If he biometric sensor is already locked out, so the current authentication attempt fails immediately
   - In this case, we report `PA2ErrorCodeBiometryNotAvailable` and this is the problem, because the error code is too vague. For example, we report such code when we cannot initialize AES/RSA cipher, so it typically means some "unexpected error".

This enhancement should change the second case and report a new error code - `PA2ErrorCodeBiometryLockout`.

hvge:
This PR implements FIX for #348 for `1.5.x` branch of SDK

### DIFF
--- a/docs/PowerAuth-SDK-for-Android.md
+++ b/docs/PowerAuth-SDK-for-Android.md
@@ -1386,6 +1386,14 @@ if (t instanceof PowerAuthErrorException) {
             android.util.Log.d(TAG,"Error code for error that occurs when pending activation is present and work with completed activation is required"); break;
         case PowerAuthErrorCodes.PA2ErrorCodeBiometryCancel:
             android.util.Log.d(TAG,"Error code for Biometry action cancel error"); break;
+        case PowerAuthErrorCodes.PA2ErrorCodeBiometryNotSupported:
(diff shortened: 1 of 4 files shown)


kober32 (approved):
OK from the code perspective, I would consider adding more detailed documentation and inline-documentation of how the biometry works o android (what lockout and not available means, etc.), some examples, and how to communicate it. 

IMO, it can raise many wrong assumptions and scenarios to a developer who is not familiar with this area (which is no wonder since it's such a mess).

PR #352, opened 2021-01-13
Android: Fix #351: Deadlock after biometric auth cancel

### LINKED ISSUE #351: Biometry deadlock when app cancels authentication
It's possible to deadlock our biometric authentication implementation in case that the application initiate authentication cancel and the error dialog from SDK is displayed. This bug affects SDKs from version `1.2.1` up to `1.5.2` running on Android 9+.

### Detailed analysis

The biometric authentication in SDK is using global lock to prevent multiple biometric authentication calls from application in the same time. This global lock is managed by two methods implemented in `BiometricAuthentication.SharedContext`:
- `startBiometricAuthentication()` - acquires lock
- `finishPendingBiometricAuthentication()` releases lock

The lock release is then initiated from `BiometricResultDispatcher` when authentication operation is properly completed or in special case, when application wants cancel the whole authentication. The cancel situation is handled with using `CancelableTask` (see dispatcher's constructor). The exact same cancelable task is suppose to be returned back to the application, so application can cancel its pending authentication request. So far, so good.

Unfortunately, the cancel code may never be executed when our own error dialog is displayed. This is due the fact, that we use `CompositeCancelableTask` to wrap dispatcher's cancelable object. This is still OK, but in case of error, the composite object forget the original cancelable interface and adds a new one, to handle dismiss of the error dialog.

### Steps to reproduce

1. Make sure that biometric sensor is temporary or permanently locked out (so, SDK will display error dialog immediately)
2. Initiate biometric authentication via `BiometricAuthenticator.authenticate()` and keep returned `ICancelable` object for later
3. Wait for at least 2 seconds (due to delay bug in Android 9 BiometricPrompt)
4. Cancel returned `ICancelable`

The result is that internal lock in `BiometricAuthentication.SharedContext` is never released, so all subsequent biometric authentication attempts will fail on simulated "cancel by user" result.

hvge:


kober32 (approved):
Looks OK

PR #358, opened 2021-03-01
Android: Fix #354: Don't use StrongBox on Google Pixel devices.

### LINKED ISSUE #354: Sudden activation loss on Google Pixel devices
We have plenty of reports from our bank customers that activation on PowerAuth based application is sporadically lost on Google Pixel 3 (and newer) devices. From the end-user perspective, the application suddenly loses its activation and switch to a default, not-activated state. After weeks of investigation, we believe that this is due to the unreliability of StrongBox implementation available on such devices. 

Since version `1.4.2`, PowerAuth mobile SDK uses Android KeyStore backed AES key to protecting its local activation data on the device. The AES-GCM with 256-bit wide key is used. Unfortunately, our implementation is very sensitive to failures, so any data decryption failure may lead to false information that there's no activation data available on the device. So, the result is that the application may end in the default state.

### Steps to reproduce

Unfortunately, there are no direct reproduction steps available for PowerAuth mobile SDK. It's challenging to reproduce this bug, so we're using a simplified [test application](https://github.com/hvge/CVE-2019-9465) that repetitive perform encrypt and decrypt operation in a similar manner to our SDK. The application is derived from the original work done for CVE-2019-9465. 

Our results clearly show that for StrongBox backed keys, the test application sooner or later fails to decrypt data. Following crash logs were collected from two instances of the test app, run in parallel:

- [pixel4-crash-1.txt](https://github.com/wultra/powerauth-mobile-sdk/files/5966632/pixel4-crash-1.txt)
- [pixel4-crash-2.txt](https://github.com/wultra/powerauth-mobile-sdk/files/5966633/pixel4-crash-2.txt)
- [pixel4-crash-3.txt](https://github.com/wultra/powerauth-mobile-sdk/files/5966634/pixel4-crash-3.txt)
- [pixel4-crash-4.txt](https://github.com/wultra/powerauth-mobile-sdk/files/5966636/pixel4-crash-4.txt)
- [pixel4-crash-5.txt](https://github.com/wultra/powerauth-mobile-sdk/files/5966637/pixel4-crash-5.txt)

All above crashes were collected on fully updated Google Pixel 4 device.

### Proposed solution

The best what we can do in this situation is to blacklist the problematic devices. So, we have to implement the following changes to SDK:

- Add a list of problematic devices (Pixel 3, 4, etc...) that should not use StrongBox backed key.
- Transparently migrate data encrypted with StrongBox key to non-StrongBox key.

The fix will be implemented for `1.4.x`, `1.5.x` and `develop` branches.

hvge:
This fix adds transparent data re-encryption in case that SDK determines that device has unreliable StrongBox implementation.


petrdvorak (approved):
Looks OK

PR #365, opened 2021-03-26
Fix #364: Clear TokenStore after activation commit

### LINKED ISSUE #364: Clear TokenStore after activation commit
We should clear TokenStore when activation is properly completed with commit. This is prevention against possible inconsistency between activation and its tokes, when status keychain suddenly lost its content, while token store keychain is preserved.

Related to #361, #354 
Will be fixed for `develop` and releases `1.4.x` and `1.5.x`

hvge:


petrdvorak (approved):
Looks OK

PR #378, opened 2021-04-19
Migrate biometric authentication to androidx.biometric

hvge:
This change removes our custom biometric authentication  and replaces the authentication backend  with `BiometricPrompt` implemented in `androidx.biometry` support lib.


TomasKypta (approved):
Looks OK

PR #383, opened 2021-04-21
iOS: Fix #367: Prepare PowerAuthAuthentication for possession+biometry in advance (1.5.x)

### LINKED ISSUE #367: Add method to return biometry related key
Currently, we have `unlockBiometryKeysWithPrompt:withBlock:` which, however, returns the entire keychain contents. We should have a similar method only for the single biometry key.

### LINKED ISSUE #376: Improve iOS error reporting
The typical question of developers using PowerAuth SDK  is:
> What does error code 9 mean?

The answer is simple if you know where to look, but that requires knowledge of SDK internals.

We have error codes defined as external integer constants, compiled as a part of `PowerAuthCore` project and that's why developers cannot find the meaning by themselves. They just see definition of external constant, but not the actual value. 

The error reporting can be easily improved just by adding default textual representations to all `NSError` objects reported from SDK. We're using `PA2MakeError()` internal function to build all errors, so the change is pretty well isolated.

hvge:
* Added missing biometry lock in unlockBiometryKeysWithPrompt method
* Added PowerAuthAuthentication static method to make possession+biometry with prompt
* Added new PA2ErrorCodeBiometryFailed error constant
* iOS: Core: Fix #376: Add text description to NSErrors

This is backport from develop to 1.5.x branch

### DIFF
--- a/docs/PowerAuth-SDK-for-iOS.md
+++ b/docs/PowerAuth-SDK-for-iOS.md
@@ -783,6 +783,32 @@ You can remove biometry related factor data used by Touch or Face ID support by 
 PowerAuthSDK.sharedInstance().removeBiometryFactor()
 ```
 
+### Fetch Biometry Credentials In Advance
+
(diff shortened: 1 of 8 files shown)


petrdvorak (approved):
Looks OK

PR #385, opened 2021-04-28
Issues/373 ios separate core lib

hvge:
This change splits iOS library into two dynamic frameworks:
- `PowerAuthCore` contains all C++ and low level Objective-C classes and links OpenSSL library under the hood.
  - The core is now a standalone library
  - All public interfaces, enums and functions, now have `PowerAuthCore` prefix
- `PowerAuth2` contains all high level SDK source files
  - There are no longer `PA2*` public classes, interfaces. All public objects, functions and enums now have `PowerAuth` prefix.
  - All previous `PA2*` interfaces are now declared as deprecated
- `PowerAuth2ForWatch` is now dynamic framework
  - All public interfaces, enums and functions, now have `PowerAuth` prefix
  - All previous `PA2*` interfaces are now declared as deprecated
- `PowerAuth2ForExtensions` is now dynamic framework
  - All public interfaces, enums and functions, now have `PowerAuth` prefix
  - All previous `PA2*` interfaces are now declared as deprecated
- All frameworks use `#import <FrameworkName/Header>` scheme for all public imports

Other changes:
- Added script `copy-shared-sources.sh` to synchronize sources between `PowerAuth2`, `PowerAuth2ForWatch` and `PowerAuth2ForExtensions` projects
- Copyright year set to 2021


kober32 (approved):
I went through the docs and didn't find any significant error other than some grammar stuff that would be better to solve in some other non-monstrous PR. The compilation of the library went OK. other than that, it's basically impossible to reasonably review it 👍🏾

PR #388, opened 2021-05-06
Android: Fix #387: Proper MainThreadExecutor singleton impl.

### LINKED ISSUE #387: Invalid synchronization in MainThreadExecutor
The synchronization is not correct in:
https://github.com/wultra/powerauth-mobile-sdk/blob/c689bc1b9c6d666254e0cdcca001b0f50bcf87f5/proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/impl/MainThreadExecutor.java#L44

The correct synchronization code in Java should use `volatile` keyword for the instance field and perform double-checked locking:
https://www.baeldung.com/java-singleton-double-checked-locking

It is also possible to avoid all such issues and simply create the instance on the field which will be initialized when the class instance gets constructed.

hvge:


romanstrobl (approved):
Yes, this fixes the problem thanks. I find the lazy approach with the holder class a bit unnecessary, because the instance will get created all the time in the app, I assume, but it is fine.

hvge:
You're right, I'll change the implementation to use a simple, early initialized value.

romanstrobl (approved):
Perfect.

PR #391, opened 2021-05-11
Core: Fix #386: Validate EC key after import

### LINKED ISSUE #386: Perform full point validation during public key import
See: https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Ar3.pdf, chapter 5.6.2.3.3.

hvge:


romanstrobl (approved):
Perfect.

hvge:
OK, I'll wait with merge for #385 to avoid possible conflicts in Xcode project file.

hvge:
> Is this change intentional?

Yes, It just prints the test data to the console. We don't need to run that test function.

PR #392, opened 2021-05-11
Android: Fix #336: Execute heavy biometry task on background thread.

### LINKED ISSUE #336: Access Android KeyStore on background thread
When asymmetric key for biometric authentication is enabled (see enhancement #332) then Android KeyStore needs more time to generate a proper RSA key-pair (two big primes must be generated). In the current implementation, this will typically freeze a main UI thread. So, SDK should internally do this on the background thread.

hvge:


TomasKypta (approved):
Looks OK

PR #394, opened 2021-05-17
Android: Fix #393: Remove classes with PA2 prefix

### LINKED ISSUE #393: Get rid of PA2 prefix in classes
There are still a few classes named with `PA2` prefix on Android platform. We should make this uniform with iOS SDK and rename it all to `PowerAuth`.

hvge:


petrdvorak (approved):
Looks OK

PR #395, opened 2021-05-17
Issues/356 ios biometry fallback

hvge:
This PR adds support for fallback from biometric authentication to device passcode.
There's also one more commit to remove warnings after upgrade to Xcode 12.5

Note that `PowerAuthKeychain.m/h` and `PowerAuthKeychainConfiguration.m/h` is available also in watchOS or app extensions projects. It's just the same source code as is for main `PowerAuth2` library.


petrdvorak (approved):
Looks OK

PR #400, opened 2021-05-24
Issues/397 unify error codes

hvge:
This PR unifies error codes returned on platforms when activation fails. The change also contains the following changes:
- update of gradle & android dependencies.
- fixed android unit tests on too old devices


kober32 (approved):
Looks OK.

PR #404, opened 2021-06-23
Apple: Fix #403: PA2WeakArray cleanup

### LINKED ISSUE #403: Crash in PA2WeakArray
Our `PA2WeakArray` implementation has an issue when an automatic array cleanup doesn't work when more than one `weak` object doesn't exist anymore. The problematic code is
```objc
[_array removeObjectAtIndex:idx]
```
The `idx` parameter may point to an invalid index.

hvge:


petrdvorak (approved):
Looks OK

PR #407, opened 2021-08-15
change sh to bash

heron-brito:
Avoid problems when importing other files using shell command "source"


hvge:
Hello @heron-brito, 
Thank you for your pull request, but please explain in more detail, why is this change important for you. You're saying that you have problems when you import files using "source", but you changed gradle scripts. That's quite odd.

Before we accept your change, we need you to file our CLA: https://forms.gle/r715RoVDoji4GD7K7 

Without it, we are unable to accept any contributions from you. While you remain the author of the change, we need you to transfer the usage right to us so that we can freely work with the software (for example, change the license later to any license we need, or use the software in our commercial software without asking your permission, etc.).

PR #416, opened 2022-03-16
Issues/412 interface nullability

hvge:
This PR fixes missing nullability annotations in public listener interfaces. 

Also I decided to deprecate `IFetchKeysStrategy` interface. It's in wrong package (networking.interfaces) and with very vague class name. The replacement is internal `IPossessionFactorEncryptionKeyProvider`.


petrdvorak (approved):
Looks OK

kober32 (approved):
👍🏻

PR #420, opened 2022-04-07
Issues/418 ios shared activation

hvge:
This PR adds functionality that allows us to share PowerAuthSDK activations across multiple apps from the same vendor.


petrdvorak (approved):
Approved

PR #428, opened 2022-05-20
Apple: Fix #421: Ignore access group config for simulator keychain (1.6.x branch)

### LINKED ISSUE #421: Ignore keychain access group in Simulator
If application specify keychain access group in `PowerAuthKeychainConfiguration` then this configuration should be ignored in our keychain implementation. This is due to fact that access groups doesn't work in simulators and if it's used then the `PowerAuthSDK` actually never save its state to simulated keychain.

This is problematic also for our own new tests, written for #418

hvge:
This is a fix for #421 backported to 1.6.x release branch.


kober32 (approved):
Looks OK 👍🏻

PR #429, opened 2022-05-20
Apple: Fix #419: Reset pending activation after failed ECIES.

### LINKED ISSUE #419: Always reset activation state when createActivation fail
The current iOS implementation suffer with an activation state inconsistency in case that `createActivation` method successfully starts the activation in low-level session and then process somehow fails during the ECIES payload encryption. We should always reset the low-level session when createActivation fails. The probability that this inconsistency happen is very low, but still we should fix it.

The implementation on Android is correct and always reset the session in case of failure.

I found this bug during the works on #418, so it will be fixed for 1.7.x release. This ticket is will cover back-port to 1.6.x branch.

hvge:
This is a fix for #419 back-ported to 1.6.x release branch.


kober32 (approved):
👍🏻

PR #432, opened 2022-05-27
Issues/188 tokens consistency

### LINKED ISSUE #188: Improve consistency between serialized local tokens and SDK state
On Android, it's possible to reset SDK's state without an actual cleanup of token store. This is possible due to optional `Context` parameter in function `PowerAuthSDK.removeActivationLocal(@Nullable Context context, boolean removeSharedBiometryKey)`. If the context is not provided, then the local token store is not properly cleaned up. Due to fact, that we're using "instance-id" to match the right data stored in the keychains, then, unfortunately, the newly created activation will reuse that old tokens.

I think that this problem cannot be simply triggered on iOS, but still, we should improve a validation whether the token cached in memory, is still valid and can be used for a header calculation.

Here's a list of proposed changes:

- **Android:** make context parameter as required in `removeActivationLocal(Context, boolean)` function.
- **Android, iOS:** serialize activation-id into local token's data.
- **Android, iOS:** improve test, whether the token header can be calculated. The instance of `PowerAuthToken` object can be cached in the memory by the application. Right now, there's a simple boolean check, whether the parent SDK object has still activation. We should compare, whether the parent SDK object has still the same activation-id.
- **Android, iOS:** After the `PowerAuthSDK` instance is created and status deserialization shows that there's no valid activation, we should automatically reset also the token store, to remove possible traces of old data.

hvge:
This PR fixes #188 - a very old issue with possible inconsistency between token store and an activation state. The goal was to add activation identifier into private token data serialized to the persistent storage. Later, we can test whether the activationID is still the same as the current activation's identifier.

To deal with an old data, the implementation on each platform automatically add missing activation-ID and re-save the token data in such case. 

At first look this is a big change, but you can review it on per-commit basis. I tried to split the fix into per-platform commits. On top of that, main iOS SDK, watchOS and extensions SDK share a lot of common token-related code.

### DIFF
--- a/docs/Migration-from-1.6-to-1.7.md
+++ b/docs/Migration-from-1.6-to-1.7.md
@@ -23,6 +23,8 @@ PowerAuth Mobile SDK in version `1.7.0` is a maintenance release that brings mul
 - Interface `IFetchKeysStrategy` is now deprecated and will be removed in the next major SDK release.
   - There's `IPossessionFactorEncryptionKeyProvider` that SDK is using internally as a replacement. If your application depends on `IFetchKeysStrategy`, then please contact us to find a proper solution for you.
 
+- Method `PowerAuthSDK.removeActivationLocal(Context, boolean)` has no longer the context parameter optional, so the Context has to be always provided.
+
 ### Other changes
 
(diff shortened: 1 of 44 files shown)


petrdvorak (approved):
Looks good

Issue #433, opened 2022-05-30
iOS app extension should not remove local token


hvge:
By adding support of activation data sharing between apps in #418 (PR #420) we have introduced a bug when iOS / tvOS app extension can create a data inconsistency in tokens. Basically, app extension should access local tokens only in read only mode, otherwise the other parties don't get the notice about the change and will use memory cached tokens.

It would be complex to add a full interprocess sharing between app extension and other apps, so it would be better to disable `removeLocalToken()` functionality for the app extension.

PR #438, opened 2022-06-09
Issues/212 group create token request

### LINKED ISSUE #212: Coalesce multiple requests for the same named token
Currently, when the mobile application requests for the same named token for twice, in the short period, then the two tokens are actually created on the server. This is not critical, because both tokens provide the same functionality, but we should avoid such situations. 

I found this quite complex to handle on the application's side, so the best way is to fix it in the SDK, and enhance the current token store implementation with following features:

- TokenStore should keep track of application requests issued to the server. If two requests asks the same token name, then the 2nd request, should wait for the token delivery.
- TokenStore should serialize signature factors used to create token. We can use this to validate the future requests and report an error, if application wants to get locally cached token with different combination of factors.

hvge:
This PR adds the following improvements to both platforms:

- Create token request grouping. So, application should issue a multiple create token requests with the same name, but only one actual token is created on the server.
- Validation of authentication factors when the token is created. If token is already created, or the create request is pending, then it's forbidden to create the same request with different factors set in `PowerAuthAuthentication`.
- Integration tests are now compatible with PowerAuth Server 1.3+ (this is the reason why this PR is so big)
- Removing local token from app extension is no longer possible (#433)

Fix #212


kober32 (approved):
Massive 🤨 But looks OK

PR #439, opened 2022-06-10
Issues/435 async operation crash

hvge:
Fixed crash in `PA2AsyncOperation` implementation. 

The crash was caused by setting `isFinished` to `YES` even if the operation was not started by the queue. That happened when the operation was canceled before it was actually started in the queue. This situation is now fixed in new `notifyStop` method. The change also adds a more locking to properties accessed by the `NSOperationQueue`.


kober32:
Ping me again pls once the https://github.com/wultra/powerauth-mobile-sdk/pull/438 will be merged...

kober32 (approved):
Looks OK

PR #444, opened 2022-06-14
Add support for custom User-Agent

hvge:
This PR adds support for application provided User-Agent added to all HTTP requests initiated from the library. The change also defines a default User-Agent string provided by PowerAuth SDK itself, if client configuration is not altered.

### DIFF
--- a/docs/PowerAuth-SDK-for-Android.md
+++ b/docs/PowerAuth-SDK-for-Android.md
@@ -1587,7 +1587,7 @@ We don't recommend implementing the `HttpRequestInterceptor` interface on your o
 
 The `PowerAuthClientConfiguration` contains `userAgent` property that allows you to set a custom value for "User-Agent" HTTP request header for all requests initiated by the library:
 
-```swift
+```java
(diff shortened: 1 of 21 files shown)


docs/PowerAuth-SDK-for-Android.md
```diff
@@ -1582,3 +1582,17 @@ final PowerAuthClientConfiguration clientConfiguration = new PowerAuthClientConf
 ```
 
 We don't recommend implementing the `HttpRequestInterceptor` interface on your own. The interface allows you to tweak the requests created in the `PowerAuthSDK` but also gives you an opportunity to break things. So, rather than create your own interceptor, try to contact us and describe what's your problem with the networking in the PowerAuth SDK. Also, keep in mind that the interface may change in the future. We can guarantee the API stability of public classes implementing this interface, but not the stability of the interface itself.
+
+### Custom User-Agent
+
+The `PowerAuthClientConfiguration` contains `userAgent` property that allows you to set a custom value for "User-Agent" HTTP request header for all requests initiated by the library:
+
+```swift
```
kober32, line 1590: Not `swift`

kober32 (approved):
Looks OK, just minor documentation issue

PR #446, opened 2022-06-20
Android: Fix #445: Android build on ARM based mac computers

### LINKED ISSUE #445: Fix Android build on M1 macs
There are several problems with build on M1 macs that should be addressed. As a part of this, we should upgrade gradle build dependencies.

hvge:
- FIxed build on ARM based mac computers
- Instrumentation testing no longer depends on parameters set in Android Studio
- Updated gradle dependencies


TomasKypta (approved):
Looks OK

PR #447, opened 2022-06-22
Issues/436 activation status refactoring

hvge:
This PR simplifies task for getting activation status. The new implementation is now using `PA2GroupedTask` / `GroupedTask` classes introduced in PR #438. On top of that, there are the following additional changes:
- On Android Platform:
  - Fixed UnblockActivationEndpoint in tests
  - Added better tests for getting activation status
  - Fixed sporadic failures in `GroupedTaskTests`
  - Fixed broken tests. The bug was introduced in PR #444
- On iOS platform
   - Custom object returned from the server together with an activation status is now a part of `PowerAuthActivationStatus` object.
   - Added new deprecations related with custom object returned from the server.
- On Both platforms:
   - Integration tests are now compatible with PA-Server 1.2.5 and supports auto-commit.

This PR can be reviewed on per-commit basis if you like.


kober32 (approved):
Kinda huge, but I went thought it, and it looks OK

PR #460, opened 2022-07-07
 PowerAuthAuthentication improvements

hvge:
This PR improves instantiation of `PowerAuthAuthentication` on both platforms. On top of that, `PowerAuthAuthentication` object has now purpose defining whether it suppose to be used for activation commit or for the signature calculation. If purpose is wrong, then the warning is reported to the debug console.


petrdvorak (approved):
Looks OK, thank you.

PR #463, opened 2022-07-14
Android: Fix #422: Detect attemtps to trick biometry sensor

### LINKED ISSUE #422: Cancel after error in android biometry
Investigate whether it's possible to catch an error in Android 9+ BiometricPrompt even when user press cancel after lockout. 

Steps to reproduce:
- Try to authenticate with biometry. 
- Try different finger / face to force system to lock biometry out. 
- When the error is displayed, press cancel button immediately.

The result is that on some devices, it's possible to press cancel and avoid increment of failed authentication counter on the server. The question is whether we can workaround this, or it's just acceptable behavior on some devices.

hvge:
This PR fixes issue on older Android devices where attacker can cancel the biometric authentication dialog after 5 subsequent failed attempts. The method presented in this PR is not 100% bulletproof, because we don't track biometric authentication attempts globally, but can catch a low skilled attackers.


petrdvorak (approved):
Looks ok

PR #468, opened 2022-08-09
Core: Fix #499: Obfuscate password stored in memory

hvge:
This PR adds simple passphrase obfuscation to our "core" password objects . The method of obfuscation is based on a simple XOR function, with using random 16 bytes as cyclic key.

The change also contains the following modifications:

- Added high level tests for `Password` (android) & `PowerAuthCorePassword` (ios, tvos)
- Removed CC7_CHECK from `Password.cpp` implementation. This has no effect on functionality, but it required due to problematic unit testing on iOS. 
- Added `Password.validatePasswordComplexity` function to Android impl.
- Changed `PowerAuthCorePassword.validatePasswordComplexity()` return value.


kober32 (approved):
Looks OK

petrdvorak (commented):
Approved

PR #470, opened 2022-08-16
Better Core Password support in SDK API

### LINKED ISSUE #465: Add better Core Password support to SDK API
We should expose `PowerAuthCorePassword` to more high-level interfaces to allow convenient work with user's credentials.

This enhancement should be implemented as non-breaking change to public interfaces, so we can release it as an update to the current, 1.7.x branch.

hvge:
This PR adds a better `PowerAuthCorePassword` (objc, swift)  / `Password` (java) support in high level SDK interfaces. The application developer now can choose between strings and the custom password implementation.


kober32 (approved):
Looks OK

PR #472, opened 2022-08-16
Android: Fix #471: Broken token data deserialization

### LINKED ISSUE #471: Exception in token data deserialization
Attempt to deserialize token data serialized in new SDK 1.7.x fails at exception due accessing wrong data at the wrong offset. The result is that application cannot access token after its restart.

hvge:


petrdvorak (approved):
Looks good

PR #473, opened 2022-08-23
Fix #464: Deployment script improvements

### LINKED ISSUE #464: Improve deployment script
The current deployment script suffers with the following problems:

1. It's damn slow, due to cocoapods
   - We have to wait about 20 minutes between releasing `PowerAuthCore` and `PowerAuth2`, because the new version is not immediately visible for 
1. It's complicated when we're deploying from non-develop branch (several switches required)

## Solution for poor release speed

We should deploy libraries in the following order:

- Run pod lib lints for all podpsecs (it's possible to trick linter for `PowerAuth2` to include `PowerAuthCore` from the local file)
- Publish android library to the local maven (to test possible failures during the real deployment)
- Publish `PowerAuthCore` pod
- Publish `PowerAuth2ForExtensions` pod
- Publish `PowerAuth2ForWatch` pod
- Publish Android library
- Wait for a shorter time
- Publish `PowerAuth2` pod
 
## Simplify script usage

- We're always deploy all platforms, so options for deploying specific platform is deprecated. 
- There should be `config.sh` specifying releasing branch and allowing automatic release from that branch
- There should be automatic detection whether the tag is already created (useful when deploy process fails on some unexpected error, like expired `pod` session)

hvge:
This PR improves deployment script speed by reordering execute of particular deployment steps. See #464 for more details.

### DIFF
--- a/scripts/android-publish-build.sh
+++ b/scripts/android-publish-build.sh
@@ -28,6 +28,8 @@ function USAGE
     echo ""
     echo "  central             Publish Android SDK to Maven Central"
     echo "  local               Publish Android SDK to local Maven cache"
+    echo "  test                Build library with all publishing artifacts"
+    echo "                      without actual publishing to repository"
     echo ""
     echo "options:"
     echo ""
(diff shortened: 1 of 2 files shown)


kober32 (approved):
👍🏻

PR #478, opened 2022-09-12
LAContext improvements

hvge:
This PR adds various improvements to LAContext usage in SDK:

- LAContext is automatically invalidated after use
- Fixed naming of PowerAuthSDK methods using LAContext
- Improved error reporting when SDK fails to fetch biometry key from the keychain
- Added "Biometry troubleshooting" section to documentation, explaining in more detail possible failures with biometry.


kober32 (approved):
Kinda huge, but looks OK.

PR #487, opened 2022-09-30
Android: Reveal activationId & fingerprint in pending activation

hvge:
This PR fixes an issue when Activation Id and Activation Fingerprint is not available after key-exchange on Android platform. 

On top of that it fixes missing comma in C++ test class. The result of the test is not affected, but test data now contains originally intended test vectors.


kober32 (approved):
Looks OK 👍

PR #489, opened 2022-10-03
iOS: Fix #488: Don't reset activation in createActivation() call

### LINKED ISSUE #488: Call to createActivation() reset already existing activation
If there's already a pending or existing activation, then another call to `PowerAuthSDK.createActivation()` does reset of the pending task or reset an existing activation.  The expected behavior is that call to createActivation does nothing to existing activation or a pending one and simply return an appropriate error.

This bug was introduced in 1.7.x release, with changes related to an interprocess locking we have implemented in iOS version.

hvge:
Don't reset activation in createActivation() if there's already some existing or pending activation.


kober32 (approved):
Looks OK

Issue #495, opened 2022-10-28
authenticateUsingBiometry doesn't increase fail counter


hvge:
If biometric authentication fails in `authenticateUsingBiometry` function, then the fake biometry factor key is not generated. The result is that fail counter is not increased on the server. This bug may affect our react-native clients, because the function is used internally to pre-authorize biometry in case of reusable authentication object is requested.

The implementation should be simple - just move the logic that determine that fake key should be generated from `signatureKeysForAuthentication` into `biometryRelatedKeyWithAuthentication` function.

---

On top of that, we may implement a few nice enhancements to the function. Right now, our react-native wrapper suffer with issue that if biometry is requested on iOS simulator, then there's no biometry dialog displayed. This is because on simulator, the direct access to the biometry protected keychain item always succeed, with no dialog displayed. On opposite to that, if `LAContext.evaluatePolicy` is used, then biometric dialog is properly displayed and the developer can decide whether it should fail or success.

So, we should reimplement `authenticateUsingBiometry` with using `LAContext` with the following benefits:

- RN wrapper will properly show biometric dialogs on iOS simulators
- We can properly handle situations mentioned in #272 and #271

PR #496, opened 2022-11-02
Improved authenticateUsingBiometry() on iOS

hvge:
This PR fixes bug #495 and enhances `authenticateUsingBiometry()` function to properly handle all possible failures that can happen during the biometric authentication.

### DIFF
--- a/docs/Migration-from-1.6-to-1.7.md
+++ b/docs/Migration-from-1.6-to-1.7.md
@@ -68,7 +68,7 @@ PowerAuth Mobile SDK in version `1.7.0` is a maintenance release that brings mul
 
 - `IOException` is no longer reported from SDK's internal networking. Now all such exceptions are wrapped into `PowerAuthErrorException` with `NETWORK_ERROR` code set.
 
-- Please read also changes introduced in [1.7.2](#changes-in-172) and [1.7.3](#changes-in-173) versions.
+- Please read also changes introduced in [1.7.2](#changes-in-172), [1.7.3](#changes-in-173) and [1.7.5](#changes-in-175) versions.
 
 ## iOS & tvOS
 
@@ -108,8 +108,6 @@ Version 1.7.3 increased minimum required iOS & tvOS deployment target to 11.0. S
 - All asynchronous methods from `PowerAuthTokenStore` protocol now returns objects conforming to `PowerAuthOperationTask` and therefore the returned operation can be canceled directly.
 
 - `PowerAuthTokenStore.cancelTask()` is now deprecated. You can cancel the returned asynchronous operation directly.
-
-- `PowerAuthSDK.authenticateUsingBiometry()` is no longer available on tvOS platform (1.7.5+)
     
 ### Other changes
 
@@ -201,3 +199,18 @@ If you still have to compile our SDK for older operating systems, then you need 
    ```
 
 If you use cocoapds for PowerAuth mobile SDK integration, then please let us know and we'll prepare a special release branch for you.
+
+## Changes in 1.7.5+
+
+### iOS
+
+The behavior of `PowerAuthSDK.authenticateUsingBiometry()` has been slightly changed and improved:
+
+- Function now properly treat biometry lockout and increase failed attempts counter on the server, See [Biometry lockout](PowerAuth-SDK-for-iOS.md#biometry-lockout) chapter.
+- Function now returns new `PowerAuthErrorCode_BiometryFallback` error code in case that user tap on the fallback button.
+- As a benefit, function now properly handles situations when the user press home or power button during the biometric authentication.
+- You can also cancel the pending authentication via the application provided `LAContext` object.
+
+### tvOS
+
+The `PowerAuthSDK.authenticateUsingBiometry()` function is no longer available on tvOS platform.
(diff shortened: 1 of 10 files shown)


docs/Migration-from-1.6-to-1.7.md
```diff
@@ -198,4 +198,19 @@ If you still have to compile our SDK for older operating systems, then you need
...
+## Changes in 1.7.5+
+
+### iOS
+
+The behavior of `PowerAuthSDK.authenticateUsingBiometry()` has been slightly changed and improved:
+
+- Function now properly treat biometry lockout and increase failed attempts counter on the server, See [Biometry lockout](PowerAuth-SDK-for-iOS.md#biometry-lockout) chapter.
+- Function now returns new `PowerAuthErrorCode_BiometryFallback` error code in case that user tap on the fallback button.
+- As a benefit, function now properly handles situations when the user press home or power button during the biometric authentication.
+- You can also cancel the pending authentication via the application provided `LAContext` object.
```
kober32, line 212: Application provided? This is weird to me, I would use "platform" or "system" provided

docs/PowerAuth-SDK-for-iOS.md
```diff
@@ -1116,6 +1116,31 @@ context.evaluatePolicy(.deviceOwnerAuthenticationWithBiometrics, localizedReason
...
+Be aware that the example above doesn't handle all quirks related to the PowerAuth protocol, so you should prefer to use `authenticateUsingBiometry()` function instead:
+
+```swift
+let context = LAContext()
+context.localizedReason = "Please authenticate with biometry"
+PowerAuthSDK.sharedInstance().authenticateUsingBiometry(withContext: context) { authentication, error in
+    guard let authentication = authentication else {
+        let nsError = error! as NSError
+        if nsError.domain == PowerAuthErrorDomain {
+            if (nsError.powerAuthErrorCode == .biometryCancel) {
```
kober32, line 1128: `If`s should be without parentheses

docs/PowerAuth-SDK-for-iOS.md
```diff
@@ -1116,6 +1116,31 @@ context.evaluatePolicy(.deviceOwnerAuthenticationWithBiometrics, localizedReason
...
 
+Be aware that the example above doesn't handle all quirks related to the PowerAuth protocol, so you should prefer to use `authenticateUsingBiometry()` function instead:
+
+```swift
+let context = LAContext()
+context.localizedReason = "Please authenticate with biometry"
+PowerAuthSDK.sharedInstance().authenticateUsingBiometry(withContext: context) { authentication, error in
+    guard let authentication = authentication else {
+        let nsError = error! as NSError
+        if nsError.domain == PowerAuthErrorDomain {
```
kober32, line 1127: I would add casting to the NSError as part of the if (`if let`). I know it cannot happen since authentication is nil in this case, but someone without the knowledge of the API doesn't know that.

proj-xcode/PowerAuth2/PowerAuthSDK.m
```diff
@@ -285,33 +285,83 @@ - (NSData*) deviceRelatedKey
...
-        *status = errSecItemNotFound;
-        return nil;
+        // If localError variable is set, then we need to report an error.
+        if (localError) {
+            if (error) { *error = localError; }
+            return nil;
+        }
+        // No error generated, so create a fake biometry key to fail on the server.
+        key = [self generateInvalidBiometricKey];
+    } else if (error) {
```
kober32, line 344: `if (error)` then success? Is this correct?

proj-xcode/PowerAuth2/PowerAuthSDK.m
```diff
@@ -1236,46 +1255,130 @@ - (void) unlockBiometryKeysWithContext:(LAContext *)context
...
+                authentication = nil;
+            }
         } else {
-            // Otherwise report an error depending on whether the operation was canceled by the user.
+            // Evaluation failed, we should investigate LAError
             authentication = nil;
-            error = PA2MakeError(status == errSecUserCanceled ? PowerAuthErrorCode_BiometryCancel : PowerAuthErrorCode_BiometryFailed, nil);
+            // Embed an original error
+            NSDictionary * errorInfo = error ? @{ NSUnderlyingErrorKey: error } : nil;
+            if ([error.domain isEqualToString:LAErrorDomain]) {
```
kober32, line 1328: On the line above, you're checking if the error is nil but here, you're just accessing it. I'm not sure how objc handles this, but shouldn't the error null check be part of the if?

docs/Migration-from-1.6-to-1.7.md
hvge: Application must provide own `LAContext` instance as a parameter to the call and then it can use `invalidate` method to cancel the pending authentication.

hvge: I'm using this wording because SDK itself create own LAContext internally, when developer use variant with prompt string at input.

proj-xcode/PowerAuth2/PowerAuthSDK.m
hvge: `error.domain` is safe because it leads to `nil`, so `[nil isEqualToString:]` is always NO. On opposite to that, if `error` is not provided (perhaps bug in iOS), then constructing a dictionary with nil value will crash. So that's why there's additional check when creating that `NSDictionary`.

proj-xcode/PowerAuth2/PowerAuthSDK.m
hvge: Yes, that's correct. This branch happens when `status == errSecSuccess`, so I would like to set error to nil. But error is optional pointer to NSError object. It would be more readable as:
```
if (status != errSecSuccess) {
   ...
} else {
   if (error) {
       *error = nil;
   }
}

```

proj-xcode/PowerAuth2/PowerAuthSDK.m
hvge: In other words, accessing `nil` object is OK, but provide `nil` to API that require some object is not OK.

kober32 (approved):
Looks OK

Issue #500, opened 2022-11-09
Allow core password copy


hvge:
We should enhance `PowerAuthCorePassword` / `core/Password` objects API with copy function that create copy of existing password into a new immutable instance. The reason behind this is that in the complex system, it's difficult to achieve isolated operations with the password and guarantee a proper passphrase destroy. We must count with the situation that some controllers may leak in memory, or may be used permanently, so the developer may use an explicit password destroy to be sure that passphrase is no longer in the memory. With the  shared password object instances, this is a quite difficult to implement.

For example, we have UI controller that is responsible to pick two PINs from the user and if both are equal, then use first one for the next operation (e.g. next UI flow). In this setup, if such controller does cleanup after its job is done, then the shared password instance is irreparably changed before the next operation complete its job. It would be nice to pass just a copy of password to the next controller and make the cleanup locally. 

### Proposed changes

- On iOS / tvOS
  - Add method for secure cleanup, that also clear an immutable password.
  - Add method that make new immutable copy of the existing password.
- On Android 
  - Add method that make new immutable copy of the existing password.

hvge:
Fixed with #501

PR #501, opened 2022-11-09
Copy Core Password

hvge:
This PR adds methods for copying core passwords into immutable objects. See associated ticket #500 for more details.


kober32 (approved):
Looks OK

Issue #503, opened 2022-11-24
Android tests: Patch address with localhost


hvge:
The Android integration tests uses default URL for servers pointing to `localhost`. This of course doesn't work, because localhost in the emulator (or if run on the device) is actually the device itself. The class loading configuration should patch this to `10.0.2.2` as we do in our [react native tests](https://github.com/wultra/react-native-powerauth-mobile-sdk/blob/develop/testapp/src/Config.ts#L43).

Issue #502, opened 2022-11-21
KeyStoreException: User authentication required


hvge:
This is a report from the customer from their developer console:

```
Exception java.security.ProviderException: Failed to generate key pair 
  at android.security.keystore.AndroidKeyStoreKeyPairGeneratorSpi.generateKeystoreKeyPair (AndroidKeyStoreKeyPairGeneratorSpi.java:514) 
  at android.security.keystore.AndroidKeyStoreKeyPairGeneratorSpi.generateKeyPair (AndroidKeyStoreKeyPairGeneratorSpi.java:470) 
  at java.security.KeyPairGenerator$Delegate.generateKeyPair (KeyPairGenerator.java:727) 
  at io.getlime.security.powerauth.biometry.impl.BiometricKeyEncryptorRsa.createRsaEncryptor (BiometricKeyEncryptorRsa.java:9) 
  at io.getlime.security.powerauth.biometry.impl.BiometricKeystore.createBiometricKeyEncryptor (BiometricKeystore.java:3) 
  at io.getlime.security.powerauth.biometry.impl.DefaultBiometricKeyEncryptorProvider.getBiometricKeyEncryptor (DefaultBiometricKeyEncryptorProvider.java:3) 
  at io.getlime.security.powerauth.biometry.BiometricAuthentication$2.run (BiometricAuthentication.java:1)
  at java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1167) 
  at java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:641) 
  at java.lang.Thread.run (Thread.java:919) 
Caused by android.security.KeyStoreException: User authentication required
  at android.security.KeyStore.getKeyStoreException (KeyStore.java:1275)
```

Device: Xiaomi Redmi Note 8Pro, Android 10

---

It looks like the user has no PIN/pattern set to unlock the phone, so attempt to create the key protected with the biometry fails. I guess that we can improve this and detect such situation before we try to create such key with the KeyStore.

PR #504, opened 2022-11-25
Use KeyGuard state in biometry evaluation

hvge:
This PR enhances how our biometry implementation detects that biometry is available for the signing on the device. Now we count also with the KeyGuard's state, so it's no longer possible to use biometry if device is locked, or it's not protected with unlock PIN, password or pattern.

The change also contains improvements in instrumentation tests, and now it's more straightforward to run tests agains PowerAuth Server running locally, on developer's computer.

The change is related to issues #502 and #503


TomasKypta (approved):
Looks OK

petrdvorak (approved):
Looks OK

PR #511, opened 2023-02-01
Apple: Fix #510: tvOS SDK is no longer required

### LINKED ISSUE #510: Do not require tvOS platform if it's not installed
In the current implementation, the build script for CocoaPods integration automatically precompile `PowerAuthCore` framework for all supported platforms (iOS, tvOS). This was great in the past while Xcode was bundled with all SDKs, but appears to be more problematic now, when all additional SDKs needs to be downloaded by the developers manually.

To improve developer experience we should detect the presence of tvOS SDK and omit such platform if SDK is not available.

hvge:
This PR changes the default behavior for tvOS platform when library is integrated with CocoaPods. The tvOS SDK is no longer enforced and if it's not installed on the development machine, then the platform is not included in the final `xcframework`.

### DIFF
--- a/PowerAuth2ForExtensions.podspec
+++ b/PowerAuth2ForExtensions.podspec
@@ -29,7 +29,7 @@ Pod::Spec.new do |s|
     s.tvos.deployment_target = '11.0'
     
     s.prepare_command = <<-CMD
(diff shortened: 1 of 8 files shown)


petrdvorak (approved):
Awesome - looks great.

Issue #512, opened 2023-02-01
Document new user info functionality


hvge:
The "user info" feature is going to be released as an update to 1.7.x release, but officially will be supported in the next major 1.8.x release. This ticket is about an official documentation for that upcoming major SDK release.

This is because we have to wait for an official support in the backend components (expected to be released together with PowerAuth Server 1.5.0), but we need that functionality available in the mobile SDK earlier than 1.8.x is released.

hvge:
Related to #506

PR #517, opened 2023-02-06
Support for /user/info endpoint

hvge:
This PR adds support for `/user/info` endpoint and improves error reported when the request encryption is required in the wrong activation state.

### DIFF
--- a/proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/ecies/EciesEncryptorFactory.java
+++ b/proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/ecies/EciesEncryptorFactory.java
@@ -62,7 +62,7 @@ public class EciesEncryptorFactory {
      */
     public @NonNull EciesEncryptor getEncryptor(@NonNull EciesEncryptorId identifier) throws PowerAuthErrorException {
         if (identifier == EciesEncryptorId.NONE) {
-            throw new PowerAuthErrorException(PowerAuthErrorCodes.WRONG_PARAMETER, "'NONE' encryptor cannot be created.");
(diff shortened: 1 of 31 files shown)


proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/ecies/EciesEncryptorFactory.java
```diff
@@ -77,23 +77,26 @@ public EciesEncryptorFactory(@NonNull Session session, @Nullable byte[] possessi
      * @throws PowerAuthErrorException if factory doesn't have {@link #mPossessionUnlockKey} but is required,
      *                                 or if low level encryptor creation fails
      */
-    private @NonNull EciesEncryptor getEncryptor(@NonNull @EciesEncryptorScope int scope, @Nullable String sharedInfo1, boolean addMetaData) throws PowerAuthErrorException {
+    private @NonNull EciesEncryptor getEncryptor(@EciesEncryptorScope int scope, @Nullable String sharedInfo1, boolean addMetaData) throws PowerAuthErrorException {
```
kober32, line 80: Do I understand correctly, that the `int` cannot be null so the attribute was unnecessary?

kober32 (changes requested):
Looks OK, but there is no documentation?

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/ecies/EciesEncryptorFactory.java
hvge: Yep, there was a warning about this.

hvge:
The documentation will be addressed in the separate ticket, see #512

kober32 (approved):
OK then

PR #518, opened 2023-02-08
Fix #505: Support for variable offline signature length

### LINKED ISSUE #505: Support offline signature with variable length
Add support for offline signatures with non-standard length. See latest changes in PowerAuth Crypto (wultra/powerauth-crypto#451)

hvge:
This PR adds support for offline signatures with variable length. The length of the signature is specified in `PowerAuthConfiguration`.


kober32 (approved):
Looks OK.

I would add a section in the docs about the impact of such change, even if they're none - like what is the backend impact (from the UX) and if there is a security to consider.

hvge:
I think that the security implications should be documented in `wultra/ powerauth-crypto`. The mobile implementation is slightly ahead of the backend components, so we can still improve this in the upcomint `1.8` release, the first with official support for this feature.

Issue #529, opened 2023-05-23
Crypto 3.2: Improve ECIES


hvge:
### Associated data and Shared Info 2

Let's define `ASSOCIATED_DATA` as data trasmitted as plaintext and included in MAC calculation. The content of `ASSOCIATED_DATA` (AD) depends on the scope of encryptor and whether data is associated with the request or the response<sup>1</sup>:

| Version | Scope       | Request | Response |
|---------|-------------|---------|----------|
| == 3.2  | APPLICATION | ByteUtils.joinStrings(VERSION, APPLICATION_KEY) | null |
| == 3.2  | ACTIVATION  | ByteUtils.joinStrings(VERSION, APPLICATION_KEY, ACTIVATION_ID) | null |
| < 3.2   | APPLICATION | null | null |
| < 3.2   | ACTIVATION  | null | null |

> Note 1: The response doesn't use AD in this protocol version 

The associated data is then applied to the updated Shared Info 2 calculation. To simplify the specification, let's define the original SH2 calculation as:

- `byte[] SH2_APP_SCOPE = Hash.sha256(APPLICATION_SECRET)`
- `byte[] SH2_ACT_SCOPE = Mac.hmacSha256(KEY_TRANSPORT, APPLICATION_SECRET)`

| Version | Scope       | Shared Info 2 calculation |
|---------|-------------|---------------------------|
| == 3.2  | APPLICATION | ByteUtils.join(SH2_APP_SCOPE, NONCE, TIMESTAMP_BYTES, KEY_EPH_PUB, ASSOCIATED_DATA) |
| == 3.2  | ACTIVATION  | ByteUtils.join(SH2_ACT_SCOPE, NONCE, TIMESTAMP_BYTES, KEY_EPH_PUB, ASSOCIATED_DATA) |
| < 3.2   | APPLICATION | SH2_APP_SCOPE |
| < 3.2   | ACTIVATION  | SH2_ACT_SCOPE |

Where:

- `TIMESTAMP_BYTES` is a byte representation of `TIMESTAMP` (see below) calculated as:
  ```java
  byte[] TIMESTAMP_BYTES = ByteUtils.encode(TIMESTAMP);
  ```
- `TIMESTAMP` is a new property of `EciesCryptogram` representing a time of the request or the response calculated as:
  ```java
  long TIMESTAMP = Time.getTimestamp();
  ```

[...]

PR #534, opened 2023-05-29
Simplified configuration

hvge:
This PR adds support for simplified cryptographic configuration.

### DIFF
--- a/include/PowerAuth/Session.h
+++ b/include/PowerAuth/Session.h
@@ -79,11 +79,11 @@ namespace powerAuth
          */
         const SessionSetup * sessionSetup() const;
         /**
-         Returns application key extracted from internal SessionSetup, or empty string
-         if session has a no valid setup.
+         Returns the application key extracted from the internal SessionSetup, or an empty string if
+         the session does not have a valid setup.
          */
         std::string applicationKey() const;
-                
+        
         /**
          Resets session into its initial state. The existing session's setup and the external encryption
          key is preserved by this call.
--- a/include/PowerAuth/Session.h
+++ b/include/PowerAuth/Session.h
@@ -78,12 +78,12 @@ namespace powerAuth
          session has a no valid setup.
          */
         const SessionSetup * sessionSetup() const;
-        
         /**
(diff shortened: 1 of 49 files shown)


include/PowerAuth/Session.h
```diff
@@ -78,12 +78,12 @@ namespace powerAuth
          session has a no valid setup.
          */
         const SessionSetup * sessionSetup() const;
-        
         /**
-         Returns value of sessionSetup()->sessionIdentifier if the setup structure is present or 0 if not.
+         Returns application key extracted from internal SessionSetup, or empty string
+         if session has a no valid setup.
```
kober32, line 83: Typo

include/PowerAuth/Session.h
```diff
@@ -78,12 +78,12 @@ namespace powerAuth
...
          */
-        cc7::U32 sessionIdentifier() const;
-        
+        std::string applicationKey() const;
+
```
kober32, line 86: Formatting

kober32 (changes requested):
Looks OK (just one typo).

But it just seems like the change is not explained enough.

If there is legacy support for RN, why is a newer server needed?
Where (as a mobile developer) can I get the new configuration?
Why the change, what's the benefit (other than just having 1 config value)?

I'm pretty sure that a lot of developers will have these questions because they will upgrade the SDK (Android studio will tell them) and they won't understand what to do and if they should nag backend developers to upgrade.

hvge:
This is just a small piece related to 1.8.0 release. The new server will be needed due to #529. I created the migration document in advance, because we know that this future release will require server 1.5.0 and newer to operate properly.

kober32 (approved):
👍🏻

hvge:
I just fixed the grammar and the formatting in that header. We'll update the documentation as a part of [other tasks](https://github.com/wultra/powerauth-mobile-sdk/milestone/11) planned for 1.8. We declare the compatibility in the compatibility matrix for each branch of SDK, so If Android developer proactively upgrade to an incompatible version, then unfortunately it will not work. There's a new status endpoint planned for server 1.5, so we can introduce an error code indicating that mobile SDK is not compatible with the server.

PR #538, opened 2023-07-28
Fix #537: Add routine to produce JWT by signing claims with private key

### LINKED ISSUE #537: Add routine to produce JWT by signing claims with private key
We already sign any data, so why not claims in order to directly produce JWT.

petrdvorak:


proj-xcode/PowerAuth2/PowerAuthSDK.m
```diff
@@ -1474,6 +1474,40 @@ - (NSData*) generateInvalidBiometricKey
...
+                                         data:claimsData
+                                     callback:^(NSData * _Nullable signature, NSError * _Nullable error) {
+        // Handle error
+        if (error) {
+            callback(nil, error);
+            return;
+        }
+        
+        // Prepare JWT Header
+        NSDictionary *header = [NSDictionary dictionaryWithObjects:@[@"ES256", @"JWT"] forKeys:@[@"alg", @"typ"]];
```
hvge, line 1493: I would prefer a more readable form here:
```
@{ @"alg": @"ES256", @"typ": @"JWT" }
```

hvge (changes requested):
Looks OK, I would just suggest you to use more modern ObjC constructs while creating constant dictionaries. 

What I miss is Android implementation.

proj-xcode/PowerAuth2/PowerAuthSDK.m
petrdvorak: I am an $#&*$, sure this looks way better...

hvge:
I have added Android implementation and integration tests.

hvge (approved):

PR #540, opened 2023-08-23
Android: Fix #539: Don't return token that is not in index.

### LINKED ISSUE #539: Android TokenStore data inconsistency
If application or mobile SDK crash in a sensitive moment between the token data is already in keychain but index is not yet written, then the token appears to exist even if it suppose not. I was able to simulate this in our unit tests when some experimental code caused a crash of SDK.

The problem is in `PowerAuthTokenStore.getTokenData()` function that doesn't verify whether the token is stored also in the index. 

This will be fixed in upcoming 1.8 and backported to 1.7.x branch.

hvge:
This PR fixes possible token data inconsistency on Android.


kober32 (approved):
Looks OK

PR #543, opened 2023-09-11
Time synchronization

hvge:
This PR adds time synchronization with the server. The PR goes to `issues/529-ecies-3.2` branch to make it simpler for review.


petrdvorak (approved):
Looks OK

PR #548, opened 2023-09-13
Android: Fix #526: Proguard rules update

### LINKED ISSUE #526: Android SDK not working with android.enableR8.fullMode=true
After enabling full mode within R8 via `android.enableR8.fullMode=true` (or by default since AGP 8+) we are getting the following error when attempting to enroll activation:

`io.getlime.security.powerauth.networking.exceptions.FailedApiException: Unable to invoke no-args constructor for class io.getlime.security.powerauth.networking.model.response.EciesEncryptedResponse. Registering an InstanceCreator with Gson for this type may fix this problem.
                                                                                                    	at io.getlime.security.powerauth.networking.client.HttpRequestHelper.buildResponseException(HttpRequestHelper.java:363)
                                                                                                    	at io.getlime.security.powerauth.networking.client.HttpRequestHelper.buildResponse(HttpRequestHelper.java:243)
                                                                                                    	at io.getlime.security.powerauth.networking.client.HttpClientTask.doInBackground(HttpClientTask.java:201)
                                                                                                    	at android.os.AsyncTask$3.call(AsyncTask.java:394)`

Can you please suggest (or even better update consumer proguard rules) so we can use full mode of R8.

hvge:
Keep whole classes instead of class members only in networking.model.**

### DIFF
--- a/proj-android/PowerAuthLibrary/proguard-rules.pro
+++ b/proj-android/PowerAuthLibrary/proguard-rules.pro
@@ -28,6 +28,6 @@
 -keepclassmembers class io.getlime.core.rest.model.** {
     <fields>;
 }
--keepclassmembers class io.getlime.security.powerauth.networking.model.** {
+-keep class io.getlime.security.powerauth.networking.model.** {
     <fields>;
 }


kober32 (approved):
👍

PR #550, opened 2023-09-19
Prepare release 1.8

hvge:
- Final migration guide to version 1.8
- iOS & tvOS min deployment target changed to 12.0
- Fixed `PowerAuthAuthentication` object immutability (related to #531)
- Fixed copy method in `PA2PrivateTokenData` object
- Final code cleanup

PR #552, opened 2023-09-20
Android: Option to disable error dialog after failed biometry

hvge:
This PR adds option to disable error dialog presented from PowerAuth SDK after biometry fail. If the error dialog is disabled, then application can detect whether failure should be displayed to the user with using new `BiometricErrorInfo` enumeration, associated to `PowerAuthErrorException`.

Other changes:

- Simplified `PowerAuthSDK.authenticateWithBiometry()`. The authenticate methods now use new `IAuthenticateWithBiometryListener` that return `PowerAuthAuthentication` in success.
- Added support for missing `BIOMETRY_NOT_ENROLLED` error code (unification with iOS)

Note that documentation doesn't contain new steps for migration to 1.8. This is covered in separate PR #550

### DIFF
--- a/docs/PowerAuth-SDK-for-Android.md
+++ b/docs/PowerAuth-SDK-for-Android.md
@@ -1772,7 +1772,7 @@ In order to obtain an encrypted biometry factor-related key for the purpose of a
 <!-- begin codetabs Kotlin Java -->
 ```kotlin
 // Authenticate user with biometry and obtain encrypted biometry factor related key.
-powerAuthSDK.authenticateUsingBiometry(context, fragment, "Sign in", "Use the biometric sensor on your device to continue", object: IAuthenticateWithBiometryListener {
+powerAuthSDK.authenticateUsingBiometrics(context, fragment, "Sign in", "Use the biometric sensor on your device to continue", object: IAuthenticateWithBiometricsListener {
     override fun onBiometricDialogCancelled(userCancel: Boolean) {
         // User cancelled the operation
     }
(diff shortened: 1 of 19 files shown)


proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/biometry/IAuthenticateWithBiometryListener.java
```diff
@@ -0,0 +1,54 @@
...
+
+import androidx.annotation.NonNull;
+import androidx.annotation.UiThread;
+import io.getlime.security.powerauth.exception.PowerAuthErrorException;
+import io.getlime.security.powerauth.sdk.PowerAuthAuthentication;
+
+/**
+ * Interface used as a callback for biometric authentication for general application use.
+ */
+public interface IAuthenticateWithBiometryListener {
```
kober32, line 27: I think we should use `biometrics` instead of biometry -> that's the feedback we received from our text proofreader, that we're using this in the wrong way.

docs/PowerAuth-SDK-for-Android.md
```diff
@@ -2567,6 +2604,13 @@ when (t) {
             PowerAuthErrorCodes.PENDING_PROTOCOL_UPGRADE -> Log.d(TAG, "The operation is temporarily unavailable, due to pending protocol upgrade.")
             PowerAuthErrorCodes.TIME_SYNCHRONIZATION -> Log.d(TAG, "Failed to synchronize time with the server.")
         }
+        // Process additional information
+        when (t.additionalInformation) {
+            BiometricErrorInfo.BIOMETRICS_FAILED_WITH_NO_VISIBLE_REASON -> {
```
kober32, line 2609: He is the usage OK

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/biometry/IAuthenticateWithBiometryListener.java
kober32: I see you're using it right in the new ENUMs, so we should generally use it properly on all new APIs.

kober32 (approved):
Looks OK. I marked one naming issue to consider changing.

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/biometry/IAuthenticateWithBiometryListener.java
hvge: OK, then it makes sense to rename the new public method to `authenticateWithBiometrics()` too

PR #555, opened 2023-09-21
Destroy authentication object

hvge:
This PR allows manual destruction of PowerAuthAuthentication object on Android. On top of that, we make sure that biometric key and other sensitive keys are also destroyed.


kober32 (approved):
Looks OK

PR #557, opened 2023-09-21
Fix #556: Document limitations of facial authentication on Android

### LINKED ISSUE #556: Document limitations of facial authentication on Android
Currently, we are responding to many questions directly to customers related to facial biometrics on Android. We should put the answer to our documentation so that we can link it.

petrdvorak:


hvge (approved):
Looks OK

PR #559, opened 2023-09-25
BiometricErrorInfo now contains localized reason of failure

hvge:
This PR changes how new `BiometricErrorInfo` is defined. Now it is a class instead of enumeration. This change allows us to embed more information than simple hint about the dialog presentation.

Other changes:

- Fixed mapping from `BiometricStatus.NOT_ENROLLED` to `PowerAuthErrorCodes.BIOMETRY_NOT_ENROLLED`
- `BiometricAuthentication.getBiometricDialogResources()` is now static (as intended)


kober32 (approved):
Looks OK

PR #564, opened 2023-11-02
Apple: Import PowerAuthCore as module in private header

hvge:
This change is important for CocoaPods integration. Import of particular header doesn't work in CocoaPods, so we have to use module import.


kober32 (approved):
looks ok

PR #567, opened 2024-01-05
Update dependencies and targetSdk to 33

### LINKED ISSUE #565: Update Android dependencies and targetSdk to 33
Every Android apk released into Google Play has to be compiled with min target 33. 
The SDK should match that to ensure correct and expected funcionality in the final app. 

Dependencies of the SDK should be updated too (ideally to their latest versions) because they most like have older targets too.

TomasKypta:
Fix #565


hvge (approved):
Looks OK.

Just for note, we should cherry-pick this also to `release/1.7.x` branch,

PR #575, opened 2024-02-19
iOS documentation fixes

### LINKED ISSUE #573: Example error in documentation
In the "Requesting Activation Status" example is a mistake

```swift
if let customObject = customObject {
}
```

This does not make any sense

---

#### Document Details

*Do not edit this section. We use it to link the issue to the appropriate document.* ⚠

* File: https://github.com/wultra/powerauth-mobile-sdk/blob/release/1.8.x/docs/PowerAuth-SDK-for-iOS.md
* Timestamp: 1695040965

### LINKED ISSUE #574: Stop using sharedInstance in documentation
We should not encourage our users to use `PowerAuthSDK.sharedInstance`.

kober32:
- Fixed #574 (stop using `sharedInstance` in the documentation)
  - this will align the documentation with the Android
  - we should not encourage our users to use `sharedInstance` anyway, because it can cause trouble later on when switching instances (end-user accounts) and is a bad pattern for modern DependencyInjection systems
  - it's a confusing point for our customers
- Fixed #573: customObject fix

### DIFF
--- a/docs/PowerAuth-SDK-for-iOS.md
+++ b/docs/PowerAuth-SDK-for-iOS.md
@@ -158,8 +158,11 @@ func application(_ application: UIApplication, didFinishLaunchingWithOptions lau
         baseEndpointUrl: "https://localhost:8080/demo-server",
         configuration: "ARDDj6EB6iAUtNm...KKEcBxbnH9bMk8Ju3K1wmjbA==")
 
-    // Configure default PowerAuthSDK instance
-    PowerAuthSDK.initSharedInstance(configuration)
+    // Create a PowerAuthSDK instance with the configuration
+    guard let powerAuth = PowerAuthSDK(configuration) else {
+        // invalid configuration
+        return false
+    }
 
     return true
 }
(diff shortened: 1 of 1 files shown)


docs/PowerAuth-SDK-for-iOS.md
```diff
@@ -2004,6 +2002,16 @@ Sometimes, you may need to develop or test your application against a service th
 ```swift
 // Set `PowerAuthClientSslNoValidationStrategy as the default client SSL certificate validation strategy`
 PowerAuthClientConfiguration.sharedInstance().sslValidationStrategy = PowerAuthClientSslNoValidationStrategy()
+
+// In case you're setting the `PowerAuthClientConfiguration` explicitly, use:
+let clientConfig = PowerAuthClientConfiguration()
+clientConfig..sslValidationStrategy = PowerAuthClientSslNoValidationStrategy()
```
hvge, line 2008: Two dots instead of one.

hvge (approved):
Looks OK, thanks for this PR. There's just a small typo in the new code block you have added.

Issue #583, opened 2024-04-23
Add possibility to tap into `PowerAuthLog`


kober32:
Scenario why we need this:

- Client distributes a test build to testers
- Testers see some issues with the app and need to provide logs
- The app gathers logs and offers the possibility to send them via email
- PowerAuth only enables logging into the console via the standard `android.util.Log` facility, which is not usable for this scenario
- Also, the client can gather errors and warnings in production as `PowerAuthLog` reports these errors via this class as well.

Solution:

Add a listener to the `PowerAuthLog` to be able to process logs by the integrator

PR #584, opened 2024-04-23
Added log listener

kober32:
Adding a log listener to be able to tap into logs and log them into file/online system.

More explanation in the issue: #583


hvge (approved):
Looks OK. Just for note, do we have plans also add similar functionality for iOS?

kober32:
Yes, I'll prepare another PR.

PR #592, opened 2024-06-05
Fix #591: Make tests compatible with latest PA server

### LINKED ISSUE #591: Tests incompatible with latest PowerAuth stack
Our integration tests are not compatible with the latest PowerAuth stack. It seems that it will be enough to declare a new PA server versions to appropriate enumerations.

hvge:
This PR fixes integration tests compatibility with the latest PowerAuth server


kober32 (approved):
👍

PR #595, opened 2024-06-05
Nullable annotation in ActivationStatus.getCustomObject()

hvge:
This PR fixes missing `@Nullable` annotation in `ActivationStatus.getCustomObject()`.


kober32 (approved):
👍

PR #608, opened 2024-07-25
Don't destroy Password when PowerAuthAuthentication is GCed

hvge:
This PR fixes potential random authentication failures caused by unwanted Password object destroy, when PowerAuthAuthentication is released by GC.


kober32 (approved):
👍

PR #617, opened 2024-09-10
Android: Fix #616: Support for devices with 16kB page size

### LINKED ISSUE #616: Add support for 16 KB devices on Android 15+
Proper 16 KB ELF alignment is necessary.
More details here https://developer.android.com/guide/practices/page-sizes#compile-16-kb-alignment

hvge:
This PR adds support for devices with 16kB page size.


TomasKypta (approved):
Looks OK.
The generated *.so files are aligned.

PR #621, opened 2024-09-13
Store biometry key for each PowerAuthSDK instance

hvge:
This PR modifies how the `PowerAuthSDK` class on Android internally stores biometry-related encryption keys. In previous versions, the key was shared between multiple `PowerAuthSDK` instances. Now, the shared key is only used if the activation has already configured a biometric factor. Once the factor using the shared key is removed, or the entire activation is removed, a per-instance key will be created during the next biometric setup.

This backward compatibility feature can be disabled in the `PowerAuthKeychainConfiguration` if required by the application. For instance, if the application uses multiple instances of `PowerAuthSDK`, it is important to disable this feature to ensure the shared key is immediately forgotten.

This PR is also back-ported to releases 1.7.x and 1.8.x (#626, #627)


petrdvorak (approved):
Looks OK

kober32 (approved):
Looks OK

PR #618, opened 2024-09-11
Update documentation

hvge:
This PR fixes multiple issues found in documentation:

- Added documentation for advanced configuration objects, such as `PowerAuthKeychainConfiguration`
- Added warning about usage of `UIDevice.current.name`
- Better activation states documentation (issue with unclear deadlock state)
- Better activation data sharing documentation
- Fixed wrong examples in instance config, still using APP_KEY, APP_SECRET, style of config

You can review it on per-commit basis.

PR #622, opened 2024-09-18
Support for OIDC activation

hvge:
This PR adds support for new OIDC activation flow. Other changes:

- There's still "CUSTOM" activation type defined for custom activations.
-  All `PowerAuthActivation` object constructors are now available with no `activationName` parameter. The new recommended way to alter activation name is to use a new builder method `setActivationName()` or `with(activationName:)`. To avoid conflicts, I'll reflect this change in documentation in separate PR #618


kober32 (approved):
Looks OK 👍

PR #634, opened 2024-10-03
Fixed public ECIES encryptor construction

hvge:
This PR fixes ECIES encryptor creation for application purposes, when `PowerAuthSDK` object is freshly instantiated.


kober32 (approved):
👍

PR #663, opened 2025-01-17
Activation data sharing docs

hvge:
This PR moves configuration of activation data sharing on iOS from App Extensions documentation to main SDK documentation. Other changes:

- Removes App Extensions documentation, because it's irrelevant now.
- Removes deprecated "manual" installation guide using git submodules.


kober32 (approved):
Looks OK

PR #664, opened 2025-01-17
Deprecate unsafeChangePassword() methods

hvge:
This PR deprecates synchronous methods that change the password locally on the device. This is due to fact that this kind of interfaces will no longer be available in Crypto4 protocol version.


kober32 (approved):
Looks OK

PR #668, opened 2025-01-24
Include debug symbols into precompiled PowerAuthCore

hvge:
This PR adds feature that allows us to include debug symbols to precompiled xcframework. The feature is also turned-on for PowerAuthCore module, integrated with CocoaPods.


kober32 (approved):
Looks OK

PR #674, opened 2025-02-12
Apple: Xcode upgrade warnings

hvge:


hvge:
Change is now part of #673

PR #675, opened 2025-02-18
Asynchronous interfaces

hvge:
This PR deprecates multiple synchronous public interfaces and introduce asynchronous interfaces that allows us future implementation details we need to implement in future SDK, supporting new protocol version.

Other notable changes:

- On Android, I introduced `PowerAuthBiometricPrompt` class that simplifies biometric authentication for various tasks. This class allows us to massively reduce combination of overloaded methods in `PowerAuthSDK` class.
- We're using `PowerAuthCoreData` (iOS) and `SecureData` (Android) now for all internal or public sensitive keys.
- On iOS, it's possible to cancel pending biometric authentication  in `authenticateUsingBiometry()` method.
- Methods for "signature calculation" now using "authorization code/header" naming.

Note that this massive change can be reviewed on per-commit basis.


proj-xcode/PowerAuth2/PowerAuthSDK.m
```diff
@@ -770,64 +770,55 @@ - (void) cancelAllPendingTasks
     return [self createActivation:activation callback:callback];
 }
 
-#pragma mark Commit
+#pragma mark Persist
 
+- (id<PowerAuthOperationTask>) persistActivationWithAuthentication:(PowerAuthAuthentication*)authentication
+                                                          callback:(void(^)(NSError * error))callback
+{
+    [self checkForValidSetup];
```
kober32, line 779: This might throw an error, but in Swift, there will be no way how to catch it, because the method won't be marked as "throwing" since it's no longer have `error:(NSError * _Nullable * _Nullable)error` as a last argument.

kober32: Just a note - I didn't try it - needs to be tested...

I wasn't able to build it via cocoapods:
```
./scripts/ios-build-sdk.sh: line 283: /Users/xxxxxxx/Library/Caches/CocoaPods/Pods/External/PowerAuthCore/2f555034c07e1c59fbf61dd7072671ba/cc7/openssl-build/version.sh: No such file or directory
```

proj-xcode/PowerAuth2/PowerAuthSDK.h
```diff
@@ -240,15 +240,53 @@
                                                           extras:(nullable NSString*)extras
                                                         callback:(nonnull void(^)(PowerAuthActivationResult * _Nullable result, NSError * _Nullable error))callback;
 
+
+/**
+ Persist activation that was created and store related data using provided authentication instance.
+ 
+ @param authentication An authentication instance specifying what factors should be stored.
+ @param callback A callback called when the process finishes - it contains an error in case of failure.
+ @exception NSException thrown in case configuration is not present.
```
kober32, line 249: Exception but the signature of the method wont be try/catch in swift.

proj-xcode/PowerAuth2/PowerAuthSDK.h
```diff
@@ -240,15 +240,53 @@
...
+ */
+- (nullable id<PowerAuthOperationTask>) persistActivationWithAuthentication:(nonnull PowerAuthAuthentication*)authentication
+                                                                   callback:(nonnull void(^)(NSError * _Nullable error))callback;
+
+/**
+ Persist activation that was created and store related data using default authentication instance setup with provided password.
+  
+ @param password Password to be used for the knowledge related authentication factor.
+ @param callback A callback called when the process finishes - it contains an error in case of failure.
+ @exception NSException thrown in case configuration is not present.
```
kober32, line 260: dtto

proj-xcode/PowerAuth2/PowerAuthSDK.h
```diff
@@ -240,15 +240,53 @@
...
+- (nullable id<PowerAuthOperationTask>) persistActivationWithPassword:(nonnull NSString*)password
+                                                             callback:(nonnull void(^)(NSError * _Nullable error))callback
+                                                                NS_SWIFT_NAME(persistActivation(withPassword:callback:));
+
+/**
+ Persist activation that was created and store related data using default authentication instance setup with provided password.
+  
+ @param password Password to be used for the knowledge related authentication factor.
+ @param callback A callback called when the process finishes - it contains an error in case of failure.
+ @exception NSException thrown in case configuration is not present.
```
kober32, line 272: dtto

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthSDK.java
```diff
@@ -913,16 +913,145 @@ public void run() {
...
+     * @param listener A callback listener called when the process finishes or fails.
+     * @return {@link ICancelable} object associated with the running HTTP request If {@code null} is returned, then the
+     *         operation failed or completed immediately.
+     */
+    public @Nullable ICancelable persisActivationWithPassword(@NonNull Context context, @NonNull Password password, @NonNull IPersistActivationListener listener) {
+        return persistActivationWithAuthentication(context, PowerAuthAuthentication.persistWithPassword(password), listener);
+    }
+
+    /**
+     * Persist activation that was created and store related data using provided password object instance.
```
kober32: `password object instance` -> copy paste error, its String in this method

kober32: Also, maybe we should encourage using the Password object rather than the String

proj-xcode/PowerAuth2/PowerAuthSDK.m
```diff
@@ -1325,6 +1326,27 @@ - (BOOL) removeBiometryFactor
     }];
 }
 
+- (id<PowerAuthOperationTask>) removeBiometryFactorWithCallback:(void (^)(NSError * _Nullable))callback
+{
+    [self checkForValidSetup];
```
kober32, line 1440: This might throw but i think that wont be "catchable" from the swift code

kober32 (changes requested):
I was unable to build it and test it in the app, but I think there is an issue in the objc/swift interop around the exception raising.

Also, I think it would be better to do it on iOS the same way as o Android - catch the exception and call the callback rather than mixing callback and exceptions together anyway.

proj-xcode/PowerAuth2/PowerAuthSDK.h
hvge: This is the way how we do it in SDK since the beginning. It's the configuration failure, so makes no sense to report it in the standard way. Maybe in the some future SDK written in swift we'll definitely redesign the error handling...

But I think we can completely remove this checks. It's there because we historically supported the static singleton like methods, where you were able to get the instance of SDK with no configuration. The static methods were deprecated in 1.9.0

hvge: I'll create a separate ticket for the cleanup.

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthSDK.java
hvge: OK, I'll create a new ticket that deprecate the "string" passwords for 1.10.0

proj-xcode/PowerAuth2/PowerAuthSDK.m
hvge: Same as above, I'll remove the checks in separate ticket.

proj-xcode/PowerAuth2/PowerAuthSDK.m
hvge: The build problem is a bug. See #683

kober32 (approved):

PR #680, opened 2025-03-11
Fix: #679 Fix link to additional OTP activation doc Android

zcgandcomp:


hvge (changes requested):
Please do not point to developers portal directly. I think the correct cross link should be `https://github.com/wultra/powerauth-crypto/blob/develop/docs/Advanced-Activation-Flows.md#additional-user-authentication-using-activation-otp`

hvge (approved):

PR #684, opened 2025-03-20
Improved demo URLs for enrollment server

### LINKED ISSUE #670: Improve base URLs in the documentation
### Task specification

In sections describing how to configure the SDK, we often use a random URL, such as `https://localhost:8080/demo-server`.

We should use something far more likely, like `https://my-server-url.com/enrollment-server`

For example here: https://developers.wultra.com/components/powerauth-mobile-sdk/1.9.x/documentation/PowerAuth-SDK-for-iOS#configuration

It's both for Android and iOS

kober32:
Fixed #670

Better demo URLs to avoid customer confusion


hvge (approved):
Looks OK

PR #685, opened 2025-03-20
Logging improvements

kober32:
### Added `PowerAuthLogToConsole`

This is because we provide a logging delegate - there might be cases when the integrator wants to process the log on its own without logging into `NSLog`

----

### `ENABLE_PA2_LOG` no longer affects `PowerAuthSystem.isInDebug`

It was undocumented and kinda unexpected that turning on the log also means that this helper returns true. We put enough obstacles to turn logs (compilation flag, enabling + verbose enabling) and enough documentation that the integrator should understand what is doing.

---

+ updated docs

### DIFF
--- a/proj-xcode/PowerAuth2/PowerAuthLog.m
+++ b/proj-xcode/PowerAuth2/PowerAuthLog.m
@@ -28,10 +28,11 @@
 #ifdef ENABLE_PA2_LOG
 static BOOL s_log_enabled = NO;
 static BOOL s_log_verbose = NO;
+static BOOL s_log_to_console = YES;
 static id<PowerAuthLogDelegate> s_log_delegate = nil;
 void PowerAuthLogImpl(NSString * format, ...)
 {
-    if (!s_log_enabled) {
+    if (!s_log_enabled || (!s_log_to_console && !s_log_delegate)) {
         return;
     }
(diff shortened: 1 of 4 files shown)


proj-xcode/PowerAuth2/PowerAuthLog.m
```diff
@@ -82,6 +85,22 @@ BOOL PowerAuthLogIsVerbose(void)
...
 }
 
+void PowerAuthLogToConsole(BOOL enabled)
+{
+#ifdef ENABLE_PA2_LOG
+    s_log_to_console = enabled;
+#endif
+}
+
+BOOL PowerAuthLogToConsoleEnabled(void)
```
hvge: I would keep consistent naming with other functions in the file:

- `PowerAuthLogToConsoleSetEnabled(BOOL)`
- `PowerAuthLogToConsoleIsEnabled(void)`

hvge (changes requested):

proj-xcode/PowerAuth2/PowerAuthLog.m
kober32: Yop 👍

hvge (approved):
Looks OK

PR #686, opened 2025-03-21
Move from "signature" to "authorization code" terminology

### LINKED ISSUE #653: Change naming from "signature" to "authorization code"
### Task specification

This task is about to change terminology from "PowerAuth symmetric signature" to "PowerAuth authorization code" in documentation and the code. We found out that the "signature" term is quite misleading for the protocol reviewers or for the developers. The typical subsequent question from the client is: "Where's the public key that validate the signature?"

This task will cover the following changes:

- Deprecate existing APIs
- Introduce new APIs for authorization code calculation
- Documentation update

hvge:
This PR changes terminology from "PowerAuth signature" to "PowerAuth authorization code". The low-level interfaces, like PowerAuthCoreSession, or C++ code, are still using "signature" terminology. I would like to update this as a part of Crypto 4 works.

### DIFF
--- a/docs/PowerAuth-SDK-for-Android.md
+++ b/docs/PowerAuth-SDK-for-Android.md
@@ -14,9 +14,9 @@
   - [Validating User Inputs](#validating-user-inputs)
 - [Requesting Device Activation Status](#requesting-activation-status)
 - [Data Signing](#data-signing)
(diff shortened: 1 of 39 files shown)


kober32 (approved):
Looks OK

PR #690, opened 2025-04-07
Xcode 16.3 compatibility

hvge:
This PR fixes SDK compilation problems on Xcode 16.3+ caused by wrong usage of `std::vector` with const non-POD type. Such C++ construct was tolerated in previous STL versions, even if it doesn't make sense, but now is forbidden.

This PR also contains the following other changes:

- Bumped C++ dialect to C++17 on both platforms
- Unified implementation of CC7_SecureClear() macro between platforms
- Fixed other C++ compilation problems, such as ambiguous usage of `byte` type
- Fixed Xcode project upgrade warnings


kober32 (approved):
Looks OK

PR #733, opened 2025-08-22
Fix #732: Crypto4: V3 activation status

### LINKED ISSUE #732: Crypto4: V3 activation status
### Task specification

_No response_

jnpsk:


src/PowerAuth/v3/ActivationServiceV3.cpp
```diff
@@ -180,7 +182,176 @@ void ActivationServiceV3::resetState()
...
+        // We don't have captured counter byte yet, so test whether the hash is OK and if yes, then keep the received byte.
+        if (hash_distance == 0) {
+            // Everything's OK, keep received byte in persistent data and suggest save the session's state.
+            pd.flags.hasAuthCodeCounterByte = 1;
+            pd.authCodeCounterByte = data.counterByte;
+        }
+        // Otherwise it's not possible to determine whether the counter's OK. We have to wait to sync counters
+        // in the next regular authorization code calculation. In this case, we must pretend that everything's OK.
+        return ActivationStatus::CounterState_OK;
+    }
```
hvge: I think we can safely ignore `hasAuthCodeCounterByte` flag and simplify the code. It was introduced in version 1.3.x, five years ago. So, we can expect that all mobile clients already have that counter byte in the persistent data.

src/PowerAuth/v3/ActivationServiceV3.cpp
```diff
@@ -180,7 +182,176 @@ void ActivationServiceV3::resetState()
...
+        .withJson(prepareRequestActivationStatus())
+        .withResponseCallback([self, context](const Request& request, const cc7::json::JsonValue& body) -> ResponseObjectPtr {
+            return self->processResponseActivationStatus(*context, body);
+        })
+        .build();
+}
+
+cc7::json::JsonValue ActivationServiceV3::prepareRequestActivationStatus()
+{
+    _status_challenge = cc7::crypto::GetRandomData(v3::STATUS_BLOB_CHALLENGE_SIZE);
```
hvge: I would rather to add this as a part of `Request` object to do not keep the challenge in the service's instance.

If you merge the latest changes from crypto4 branch, then there's possibility to assign `cc7::crypto::Parameter` to the `Request` object. This PR inspired me to add this possibility.

In the `RequestBuilder`, use
```cpp
.withCustomParameter(cc7::crypto::Parameter::copy(challenge))
```
and then, you can retrieve the stored value:
```
request.getCustomParameter().asByteRange()
```

hvge (changes requested):
Looks great! I just suggested a couple of small improvements in my comments.

src/PowerAuth/v3/ActivationServiceV3.cpp
jnpsk: Ok, removed. Thanks for explanation!

hvge (approved):

PR #741, opened 2025-08-29
Fix #716: Crypto4: Add V3 auth code calculator

### LINKED ISSUE #716: Crypto4: Add V3 auth code calculator
### Task specification

Add implementation for service calculating authorization codes for V3 protocol.

Depends on #652

jnpsk:


src/PowerAuth/v3/AuthenticationServiceV3.cpp
```diff
@@ -0,0 +1,158 @@
...
+{
+    LOCK_GUARD();
+    
+    if (_session_data->hasRegistrationData()) {
+        throw Exception(EC_WrongActivationState, "Authentication header calculation is not allowed during activation registration");
+    }
+    
+    if (_session_data->hasUpgradeData()) {
+        throw Exception(EC_WrongActivationState, "Authentication header calculation is not allowed during pending protocol upgrade");
+    }
```
hvge: I think there will be exception in the upgrade process itself. The request that initiate the upgrade process is signed with V3 signature while `hasUpgradeData()` will be true at the time of signature computation. So, I recommend you to also test `auth_data.allowedInUpgrade` flag.

hvge (approved):
Looks OK. I found just a minor issue that will be problematic later, when we'll implement activation upgrade.

PR #746, opened 2025-09-05
Added Task and GetActivationStatusTask classes

hvge:
This PR adds base Task class that allows us execute more complex operations, composed from multiple HTTP requests in C++. The request also implements task for getting activation status.


jnpsk (approved):
👏

PR #751, opened 2025-10-02
Fix #750: Modify Request class to allow protocol upgrade scenario

### LINKED ISSUE #750: Modify Request class to allow protocol upgrade scenario
### Task specification

To support the protocol upgrade scenario, we need to support following.
- `RequestBuilder` uses services available within the context, we should allow overriding service instances so we can combine v4 encryption with v3 authentication.
- By default, encryption header is omitted for authenticated request. For the protocol upgrade, we need to enforce both headers.  Implement `EndpointSpec` flag to enforce encryption header for authenticated request.

jnpsk:


hvge (approved):
Looks OK

PR #754, opened 2025-10-08
Fix #752: Add V4 Protocol Upgrade Task

### LINKED ISSUE #752: Add V4 Protocol Upgrade Task
### Task specification

_No response_

jnpsk:


src/PowerAuth/Context.cpp
```diff
@@ -230,6 +254,7 @@ void Context::createServices(bool initial_setup, ConstPowerAuthSpecPtr specifica
         // V4
         _shared_secret = SharedSecret::getInstance(_specification->sharedSecret());
         _key_provider = std::make_shared<v4::KeyProviderV4>(self);
+        _key_provider->asService()->restoreSensitiveData();
```
jnpsk, line 303: The `restoreSensitiveData()` call was extracted from the constructor of the `KeyProviderV4`, as in some case it may result in attempt to create `SecretKeysV4`, which constructor expects reference to the (not yet existing) `KeyProviderV4` instance.

src/PowerAuth/v4/AeadEncryptorFactory.cpp
```diff
@@ -90,7 +90,7 @@ bool AeadEncryptorFactory::hasTemporaryKey(EncryptorScope scope)
     checkNotDestroyed();
     auto& ki = keyInfo(scope);
     if (ki.isValid()) {
-        if (!ki.isExpired(_time_service->currentTimeMillis())) {
+        if (!ki.isExpired(_time_service->currentTime())) {
```
jnpsk, line 93: Just a fix, as before we compared milliseconds to seconds.

src/PowerAuth/v4/KeyProviderV4.cpp
```diff
@@ -274,6 +282,39 @@ std::unique_ptr<PersistentData> KeyProviderV4::createPDFromSecretKeys(SecretKeys
...
+    auto spec = PowerAuthSpec::specForAlgorithm(_configuration->algorithm());
+    const auto& ud = _session_data->upgradeData().v4();
+    const auto& pd_v3 = _session_data->persistentData().v3();
+    
+    // create new V4 persistent data
+    auto pd = std::make_unique<PersistentData::V4>();
+    
+    pd->algorithmId = spec->algorithmId();
+    pd->activationId = pd_v3.activationId;
+    pd->authCodeCounterByte = pd_v3.authCodeCounterByte;
```
jnpsk, line 296: The existing counter will remain.

src/PowerAuth/v4/AeadEncryptorFactory.cpp
hvge: Good catch!

src/PowerAuth/task/ProtocolUpgradeTask.cpp
```diff
@@ -0,0 +1,205 @@
...
+{
+    Task::onTaskStart();
+    
+    switch (_session_data->getCurrentProtocolVersion()) {
+        case Version_V3:
+            startProtocolUpgrade();
+            break;
+            
+        case Version_V4:
+            /// Local activation is alredy on V4, check the server state and send the confirm request if expected.
```
hvge: `alredy` - typo

src/PowerAuth/task/ProtocolUpgradeTask.cpp
```diff
@@ -0,0 +1,205 @@
...
+    }
+}
+
+void ProtocolUpgradeTask::startProtocolUpgrade()
+{
+    LOCK_GUARD();
+    auto current_context = lockContext();
+    
+    auto &pd = _session_data->persistentData().v3();
+    pd.flags.pendingUpgradeVersion = Version_V4;
```
hvge: I would rather not change this property, because its value is persisted to local storage. It may be problematic on iOS, when two apps use activation data sharing and 2nd application is still on older SDK, not supporting V4.

src/PowerAuth/task/ProtocolUpgradeTask.h
```diff
@@ -0,0 +1,69 @@
...
+
+namespace powerAuth {
+
+class ProtocolUpgradeTask : public Task
+{
+public:
+    ProtocolUpgradeTask(const ContextPtr& context, const PasswordPtr& password, const cc7::ByteRange& new_biometry_kek);
+    
+protected:
+    void onTaskStart() override;
```
hvge, line 30: I think we should also override `onTaskEnd()` and call the `resetState()` function. This change should guarantee task cleanup in case of unexpected failure.

src/PowerAuth/task/ProtocolUpgradeTask.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+void ProtocolUpgradeTask::onRequestFailure(const Request &request)
+{
+    switch (request.getParentTaskTag()) {
+        case START_UPGRADE:
+            resetState();
+            break;
+        
+        case CONFIRM_UPGRADE:
+            fetchActivationStatus();
```
hvge: We have to limit number of attempts here, otherwise we can end in infinite loop. For example, if confirm fails and status not.

src/PowerAuth/Context.cpp
```diff
@@ -216,6 +222,24 @@ ContextPtr Context::getInstance(ConfigurationPtr configuration)
...
+{
+}
+
+ContextPtr Context::getTargetAlgorithmInstance(const ContextPtr& primary_context)
+{
+    auto config = primary_context->configuration();
+    auto spec = PowerAuthSpec::specForAlgorithm(config.algorithm());
+    
+    auto target_context = std::make_shared<Context>(*primary_context);
+    target_context->createServices(false, spec);
```
hvge: This is a bit fishy for me. If created like this, then `sessionData` and `timeService` are not initialized in the target context. I wonder how it works, because `TimeService`  is definitely required in `AeadEncryptorFactory` and its encryptors.

src/PowerAuth/model/UpgradeData.h
```diff
@@ -0,0 +1,56 @@
...
+        cc7::crypto::KeyPairPtr deviceKeyPair;
+        cc7::crypto::PublicKeyPtr serverPublicKey;
+        
+        ISharedSecretPtr sharedSecretAlgorithm;
+        SharedSecretContextPtr sharedSecretContext;
+        cc7::ByteArray calculatedSharedSecret;
+        
+        cc7::ByteArray authCodeCounterData;
+        
+        std::shared_ptr<Context> context;
```
hvge: I would rather to move the ownership of the temporary target context to `Context` object itself. I have several reasons for that:

- If we fix the missing SessionData reference in the target Context (see my other comment), then there will be a cyclic reference loop. Check this chain:
  ```
   Context (main) -> SessionData -> UpgradeData -> Context (target) -> SessionData (again)
  ```
- It would be nice to reuse the services created in the target context, when the upgrade is done. It will save at least one call for getting the temporary encryption key again.

hvge (changes requested):
This is a good start, but I found several problems with this implementation. See my comments.

src/PowerAuth/task/ProtocolUpgradeTask.cpp
jnpsk: If the first confirm fails, an activation status is fetched. If the activation fetch succeeds, the confirm is sent once again. The is no trial loop in the task.
However, the `GetActivationStatusTask` was modified, so every time the status blob contains the waiting-for-confirm flag, the confirm protocol upgrade is sent, so the protocol is eventually confirmed.

src/PowerAuth/Context.cpp
jnpsk: It is done due to the copy constructor of the Context. The `sessionData` and the `timeService` is shared.

src/PowerAuth/model/UpgradeData.h
jnpsk: Indeed, there is cyclic reference. I liked the idea of having the upgrade context inside a struct with other upgrade data. As we do not rely on RAII, we manually break the loop and destroy used resources after the upgrade.
However, the approach of reusing the services after the upgrade procedure would be a nice touch. I will reconsider this.

src/PowerAuth/Context.cpp
hvge: Great, I overlooked the copy-construction. Thanks.

hvge (approved):
Looks OK

PR #755, opened 2025-10-09
Fix #745: Crypto4: Implement fetch user info data

### LINKED ISSUE #745: Crypto4: Implement fetch user info data
### Task specification

This ticket is about plugging back functionality that fetch data about user from the server. 

Sub-tasks:
- Add method to `IActivationService` for fetching user info data. Method returns `Request` object. 
- Enhance `SessionData` class to keep last fetched, in form of parsed `cc7::json::JsonValue` object.
- Implement `fetchUserInfo()` and `lastFetchedUserInfo`  in all involved classes and objects: 
  - `Session.cpp` - core implementation
     - is also responsible for managing the last fetched user info object.  
     - make sure `lastFetchedUserInfo` is reset when activation is removed
  -  `PowerAuthCoreSession.mm` - Objective-C wrapper on top of C++ core
  -  `PowerAuthSDK.m` - public interface

jnpsk:


proj-xcode/PowerAuth2/PowerAuthSDK.m
```diff
@@ -1766,35 +1762,28 @@ @implementation PowerAuthSDK (UserInfo)
 - (PowerAuthUserInfo*) lastFetchedUserInfo
 {
     [_lock lock];
```
hvge: I think the lock - unlock is no longer required, because you moved storage to C++ session, which is already thread safe.

proj-xcode/PowerAuthCore/PowerAuthCoreSession.h
```diff
@@ -146,6 +146,8 @@
  */
 @property (nonatomic, assign, readonly) PowerAuthCoreProtocolVersion protocolVersion;
 
+@property (nonatomic, assign, readonly, nonnull) NSDictionary<NSString*, NSObject*>* lastUserInfo;
```
hvge: Missing documentation

proj-xcode/PowerAuthCore/PowerAuthCoreSession.h
```diff
@@ -235,6 +237,10 @@
 - (nullable PowerAuthCoreRequest*) removeActivationWithCredentials:(nonnull PowerAuthCoreCredentials*)credentials
                                                              error:(NSError*_Nullable*_Nullable)error;
 
+#pragma mark - User info
+
+- (nullable PowerAuthCoreRequest*) fetchUserInfo:(NSError*_Nullable*_Nullable)error;
```
hvge, line 255: Missing documentation.

src/PowerAuth/model/SessionData.cpp
```diff
@@ -70,6 +70,7 @@ void SessionData::setRegistrationData(RegistrationDataPtr &ptr)
     _modified = _pd != nullptr;
     _rd = std::move(ptr);
     _pd = nullptr;
+    _user_info = cc7::json::JsonValue::object();
```
hvge: I think we should use JsonValue's constructor with no parameters here. It will create "NaT" type of JsonValue and that's by default converted to `nil` at ObjC level. Otherwise this will be a breaking API change. 

To test this situation, modify beginning of `- (void) testUserInfo()` to
```objc
- (void) testUserInfo
{
    CHECK_TEST_CONFIG();
    
    //
    // This test validates that the User Info claims are stored correctly after activation
    // and properly updated after fetching new User Info from the server.
    //
    
    XCTAssertNil(_sdk.lastFetchedUserInfo);
    
    PowerAuthSdkActivation * activation = [_helper createActivation:YES];
```

The current implementation crash, because there's unhandled C++ exception.

src/PowerAuth/model/SessionData.cpp
```diff
@@ -100,6 +101,7 @@ void SessionData::resetSessionData()
     _modified = _pd != nullptr;
     _rd = nullptr;
     _pd = nullptr;
+    _user_info = cc7::json::JsonValue::object();
```
hvge: Same here, use `JsonValue()` constructor.

src/PowerAuth/v4/ActivationServiceV4.cpp
```diff
@@ -127,6 +127,10 @@ ResponseObjectPtr ActivationServiceV4::processResponseActivationData(Context& co
     rd.sharedSecretContext = nullptr;
     rd.sharedSecretAlgorithm = nullptr;
     
+    // Store received User Info into Session Data
+    auto user_info = L1_data.findValueAtPath("userInfo");
+    _session_data->setUserInfo(user_info ? *user_info : cc7::json::JsonValue::object());
```
hvge: Same here, use `JsonValue()`

src/PowerAuth/v3/ActivationServiceV3.cpp
```diff
@@ -118,6 +118,10 @@ ResponseObjectPtr ActivationServiceV3::processResponseActivationData(Context& co
     rd.authCodeCounterData      = ctr_data;
     rd.serverPublicKey          = server_public_key;
     
+    // Store received User Info into Session Data
+    auto user_info = L1_data.findValueAtPath("userInfo");
+    _session_data->setUserInfo(user_info ? *user_info : cc7::json::JsonValue::object());
```
hvge: Same here, use `JsonValue()`

proj-xcode/PowerAuthCore/PowerAuthCoreSession.mm
```diff
@@ -192,6 +192,12 @@ - (PowerAuthCoreProtocolVersion) protocolVersion
     return (PowerAuthCoreProtocolVersion) _session->getProtocolVersion();
 }
 
+- (NSDictionary<NSString*, NSObject*>*) lastUserInfo
+{
+    [self requireReadAccess:nil];
+    return cc7::objc::JsonValueToObjC(_session->lastUserInfo());
```
hvge, line 198: Here's unhandled exception in case activation is not present. I think we can simplify this and annotate `lastUserInfo()` method with `noexcept`.

src/PowerAuth/Session.cpp
```diff
@@ -226,6 +226,20 @@ RequestPtr Session::removeBiometricFactor()
...
+    checkActivationData();
+    return _context->activationService().fetchUserInfo();
+}
+
+const cc7::json::JsonValue& Session::lastUserInfo() const
+{
+    LOCK_GUARD();
+    checkActivationData();
+    return _context->sessionData().getUserInfo();
+}
```
hvge: This is dangerous because the method returns a reference to internal data. You’re trying to avoid a race condition with `GUARD_LOCK()`, but then the reference is returned and the object is accessed outside of the thread-safe block.

Let’s mark the method as `noexcept` and return a copy of the user info object instead.

hvge (changes requested):
I found several problems in the code:

- API is not compatible with 1.9.x
- Access to cached user info data is not thread safe.

src/PowerAuth/model/SessionData.cpp
jnpsk: Good catch, thanks for the suggestion!

hvge (approved):
Looks OK, thanks!

PR #777, opened 2025-10-23
Notify C++ Request about HTTP request failure

hvge:
This PR fixes several problems in crypt4 branch:

- Fixes cancelation of time synchronization task in TimeService
- Fixes error propagation in HTTP request processing
- Fixes C++ Request cancelation 
- Server status fetch now uses v3 or v4 endpoint, depending on the current protocol version.

Other improvements:

- Added support for simulating HTTP request failures.
- Added `BaseTestWithActivation` as most generic base class for integration tests based on activation


jnpsk (approved):
Thanks!

PR #783, opened 2025-11-05
Fix #753: Add ObjC Wrapper for V4 Protocol Upgrade Task

### LINKED ISSUE #753: Add ObjC Wrapper for V4 Protocol Upgrade Task
### Task specification

_No response_

jnpsk:


proj-xcode/PowerAuth2/PowerAuthSDK.m
```diff
@@ -901,6 +903,85 @@ - (PowerAuthActivationStatus*) lastFetchedActivationStatus
...
+    
+    id<PowerAuthOperationTask> task = [_protocolUpgradeTask createChildTask:callback];
+    if (!task) {
+        PowerAuthCoreData * biometryKek = nil;
+        if (self.hasBiometryFactor) {
+            if (newBiometryKek) {
+                biometryKek = newBiometryKek;
+            } else {
+                biometryKek = [_sessionInterface readTaskWithSession:^PowerAuthCoreData* _Nullable(PowerAuthCoreSession* session, NSError** error) {
+                    return [PowerAuthCoreSession generateFactorKekForProtocolVersion:PowerAuthCoreProtocolVersion_V4 error:error];
```
hvge: `generateFactorKekForProtocolVersion` class method doesn't need session locking. The call is similar to static method call, with no `this` and therefore no data locking is needed.

Also, I would rather not swallow the error potentially reported from the function.

proj-xcode/PowerAuth2/PowerAuthSDK.h
```diff
@@ -348,6 +348,44 @@
...
+ @param password Required core password instance used to authenticate the protocol upgrade start request.
+ @param newBiometryKek Optional parameter. If a biometry factor is configured in the current protocol version
+ and no new biometry KEK is provided, one will be automatically generated for the upgraded protocol.
+ @param callback A callback called when the upgrade task finishes.
+ @return Protocol upgrade task instance.
+ */
+- (nullable id<PowerAuthOperationTask>) startProtocolUpgradeWithCorePassword:(nonnull PowerAuthCorePassword*)password
+                                                          withNewBiometryKek:(nullable PowerAuthCoreData*)newBiometryKek
+                                                                    callback:(nonnull void(^)(id _Nullable result, NSError * _Nullable error))callback
+                            NS_SWIFT_NAME(startProtocolUpgradeWithCorePassword(password:withNewBiometryKek:callback:));
```
hvge: Please remove "core" terminology from Swift interface to be consistent with the rest of PowerAuthSDK API. The best example is `validatePassword(password:callback)` swift method, that's actually mapped to two different Objective-C messages:
```objc
// Core password version
- (nullable id<PowerAuthOperationTask>) validateCorePassword:(nonnull PowerAuthCorePassword*)password
                                                    callback:(nonnull void(^)(NSError * _Nullable error))callback
                            NS_SWIFT_NAME(validatePassword(password:callback:));

// String version
- (nullable id<PowerAuthOperationTask>) validatePassword:(nonnull NSString*)password
                                                callback:(nonnull void(^)(NSError * _Nullable error))callback
                            NS_SWIFT_NAME(validatePassword(password:callback:));
```

So, we should also intoduce a "NSString" variant of `startProtocolUpgrade()` method, to be consistent with the rest of API.

Btw,  I'm using `ContentView.swift` available in `PowerAuth2TestHostApp` that allows me to simulate how developers will see our API from Swift.

proj-xcode/PowerAuth2/PowerAuthSDK.h
```diff
@@ -348,6 +348,44 @@
...
+ Confirm the protocol upgrade.
+ Internally a new protocol upgrade task is initiated without passing a password, and so it is only allowed
+ to confirm the protocol upgrade. This is used when the upgrade task was already started, the protocol
+ was updated locally, but the final confirmation step has not yet been completed.
+ 
+ @param callback A callback called when the upgrade task finishes.
+ @return Protocol upgrade task instance.
+ */
+- (nullable id<PowerAuthOperationTask>) confirmProtocolUpgrade:(nonnull void(^)(id _Nullable result, NSError * _Nullable error))callback
+                            NS_SWIFT_NAME(confirmProtocolUpgrade(callback:));
```
hvge: We don't need to expose this to the application. If upgrade task fails and client is able to recover from the situation, then we can do the recovery automatically as a part of getting activation status.

proj-xcode/PowerAuth2/PowerAuthSDK.m
```diff
@@ -901,6 +903,85 @@ - (PowerAuthActivationStatus*) lastFetchedActivationStatus
...
 
+#pragma mark - Protocol upgrade
+
+- (id<PowerAuthOperationTask>) startProtocolUpgradeWithCorePassword:(PowerAuthCorePassword*)password
+                                                 withNewBiometryKek:(PowerAuthCoreData*)newBiometryKek
+                                                           callback:(void(^)(id status, NSError * error))callback
+{
+    [_lock lock];
+    
+    id<PowerAuthOperationTask> task = [_protocolUpgradeTask createChildTask:callback];
```
hvge: I think this is not a good idea. The grouping task with child cancelables is ideal for getting activation status, where's a high chance that application, by mistake, call for the status for twice, from the different parts of the code. This really happens, because applications are complex. 

The upgrade is not such task. It's authenticated and therefore require user's attention, and is basically very similar to activation create. So, If there's pending upgrade, then another call to `startProtocolUpgrade()` should fail. 

What's important and is missing in the function, is proper IPC notify about the upgrade start. Look for `startExternalPendingOperation` in `createActivation` method as an example.

proj-xcode/PowerAuth2/PowerAuthSDK.h
```diff
@@ -348,6 +348,44 @@
...
+ Start the protocol upgrade process without specifying a new biometry KEK.
+ If a biometry factor is configured in the current protocol version,
+ a new biometry KEK  will be automatically generated for the upgraded protocol.
+ 
+ @param password Required core password instance used to authenticate the protocol upgrade start request.
+ @param callback A callback called when the upgrade task finishes.
+ @return Protocol upgrade task instance.
+ */
+- (nullable id<PowerAuthOperationTask>) startProtocolUpgradeWithCorePassword:(nonnull PowerAuthCorePassword*)password
+                                                                    callback:(nonnull void(^)(id _Nullable result, NSError * _Nullable error))callback
```
hvge: I think that `callback` should take a new object in the result parameter - `PowerAuthUpgradeResult`. My idea is to make this object similar to `PowerAuthActivationResult`.

 In the first version, it may contain `activationFingerprint`, just like activation result has. We'll extend it later by adding more post-upgrade information, such as vault encryption keys, etc.

It might contain an information whether the upgrade has been completed correctly, including confirmation.

proj-xcode/PowerAuth2/private/PA2ProtocolUpgradeTask.h
```diff
@@ -0,0 +1,95 @@
...
+- (void) startProtocolUpgradeTask:(PA2ProtocolUpgradeTask*)task
+             didFinishedWithError:(NSError*)error
+            newBiometryKekToStore:(PowerAuthCoreData*)newBiometryKek;
+
+@end
+
+/**
+ The `PA2ProtocolUpgradeTask` class implements protocol upgrade procedure.
+ */
+@interface PA2ProtocolUpgradeTask : PA2GroupedTask<id>
```
hvge: I think this interface is not necessary . We can handle all the task logic in C++ core.

proj-xcode/PowerAuth2/private/PA2CoreHttpClient.h
```diff
@@ -86,7 +86,8 @@
 /// happens while sending the request, so no data is received on the server.
 /// - Parameters:
 ///   - relativePath: Relative path. If you use `nil` or `"*"`, then any request will fail.
-+ (void) setNextRequestNetworkFailureOnSend:(nullable NSString*)relativePath;
++ (void) setNextRequestNetworkFailureOnSend:(nullable NSString*)relativePath
+                                repeatCount:(NSInteger)count;
```
hvge, line 90: Nice improvement, I was thinking about this, but was lazy to implement it :)

hvge (changes requested):
I found problems that should be resolved.

proj-xcode/PowerAuth2/PowerAuthSDK.h
```diff
@@ -348,6 +349,59 @@
...
+ @param password Required core password instance used to authenticate the protocol upgrade start.
+ @param newBiometryKek Optional parameter. If a biometry factor is configured in the current protocol version
+ and no new biometry KEK is provided, one will be automatically generated for the upgraded protocol.
+ @param callback A callback called when the upgrade task finishes.
+ @return Protocol upgrade task instance.
+ */
+- (nullable id<PowerAuthOperationTask>) startProtocolUpgradeWithCorePassword:(nonnull PowerAuthCorePassword*)password
+                                                          withNewBiometryKek:(nullable PowerAuthCoreData*)newBiometryKek
+                                                                    callback:(nonnull void(^)(PowerAuthProtocolUpgradeResult * _Nullable result, NSError * _Nullable error))callback
+                            NS_SWIFT_NAME(startProtocolUpgrade(password:withNewBiometryKek:callback:));
```
hvge: Please rename `newBiometryKek` to `customBiometryKek` to match naming used in `addBiometryFactor()` function.

proj-xcode/PowerAuthCore/PowerAuthCoreProtocolUpgradeResult.mm
```diff
@@ -0,0 +1,34 @@
...
+@implementation PowerAuthCoreProtocolUpgradeResult
+
+- (instancetype) initWithProtocolUpgradeResult:(const powerAuth::ProtocolUpgradeResult&)protocolUpgradeResult
+{
+    self = [super init];
+    if (self) {
+        _activationStatusFetchRequired = protocolUpgradeResult.isPendingUpgradeConfirm();
+        _activationFingerprint = protocolUpgradeResult.activationFingerprint().empty()
+            ? nil
+            : cc7::objc::CopyToNSString(protocolUpgradeResult.activationFingerprint());
```
hvge: There's `CopyToNullableNSString()` helper function with the same functionality.

proj-xcode/PowerAuth2IntegrationTests/PowerAuthSDKBaseTests.m
```diff
@@ -2400,4 +2400,364 @@ - (void) testFailedStatusFetch
...
+
+- (void) testProtocolUpgradeEc
+{
+    CHECK_TEST_CONFIG();
+    
+    //
+    // Test successful upgrade from V3 to V4 protocol,
+    // with target algorithm suite EC_P384.
+    //
+    const PowerAuthAlgorithm targetAlgorithm = PowerAuthAlgorithm_EC_P384;
```
hvge: Please use `self.powerAuthAlgorithm` that contains target algorithm for the current test. The goal is to test upgrade for all possible target V4 algorithms, in all success or failure scenarios. The helper's prepare method should automatically create V3 activation and then re-create SDK to the target algorithm.

In case the current algorithms is "legacy", you can skip the test execution (or disable in the scheme). Additionally, you can add one more test case that will check whether upgrade is unavailable if the target algorithm is "legacy".

include/PowerAuth/ProtocolUpgradeResult.h
```diff
@@ -0,0 +1,67 @@
...
+    
+    /**
+     * Creates a `ProtocolUpgradeResult` indicating that the upgrade confirm is pending.
+     */
+    static std::shared_ptr<ProtocolUpgradeResult> UpgradeConfirmPending();
+    /**
+     * Creates a `ProtocolUpgradeResult` indicating that the upgrade has been confirmed.
+     * @param activation_fingerprint Upgraded activation fingerprint.
+     */
+    static std::shared_ptr<ProtocolUpgradeResult> UpgradeConfirmed(const std::string& activation_fingerprint);
```
hvge: Just nitpick, I'm using lowercase letter at the beginning of method name

src/PowerAuth/model/PersistentData.h
```diff
@@ -108,6 +108,15 @@ class PersistentData
...
         cc7::ByteArray  cDevicePrivateKey;
+        
+        struct _Flags {
+            /// True if the session is currently in a protocol upgrade process.
+            cc7::U32 pendingProtocolUpgrade : 1;
+        };
+        union {
+            _Flags      flags;
+            cc7::U32    flagsU32;
+        };
```
hvge, line 119: There's missing serialization / deserialization of flagsU32 value in implementation. We should persist this information, because application may crash during the process of upgrade. If flag is not persistent, then the behavior after app restart would be less deterministic.

You can use `reCreateSdkInstanceWithConfiguration` in tests to simulate app restart.

proj-xcode/PowerAuth2IntegrationTests/PowerAuthSdkTestHelper.m
```diff
@@ -485,6 +485,72 @@ - (void) cleanup
...
+    
+    if (biometryKek) {
+        XCTAssertFalse(_sdk.hasBiometryFactor);
+        [AsyncHelper synchronizeAsynchronousBlock:^(AsyncHelper *waiting) {
+            [_sdk addBiometryFactorWithCorePassword:_currentActivation.credentials.password customBiometryKek:biometryKek callback:^(NSError * _Nullable error) {
+                XCTAssertNil(error);
+                [waiting reportCompletion:nil];
+            }];
+        }];
+    }
```
hvge: There's method `createActivationWithFlags` where you can use `TestActivationFlags_PersistWithBiometry` or `TestActivationFlags_PersistWithFakeBiometry` to create activation automatically with biometry key. 

The option with fake biometry allows you to also test biometric authentication with pre-defined biometric KEK. The appropriate key is available in `PowerAuthSdkActivation` and `biometryCredentials` property.

hvge (changes requested):
There's missing serialization of pending upgrade flag and also we should improve integration tests a bit.

hvge (approved):
I've pushed a small improvements to this PR:

1. Fix for error propagation in `PA2CoreHttpClient`, when request peparation fails.
1. Fix in `PowerAuthTestServerAPI` for getting activation status
1. Improvements in upgrade tests

Looks OK now.

PR #785, opened 2025-11-07
Fix #784: Crypto4: Move last fetched activation status to session data

### LINKED ISSUE #784: Crypto4: Move last fetched activation status to session data
### Task specification

Currently the last fetched activation status object is cached in the PowerAuthSDK class. Move the ownership of the cached object to the SessionData, similarly as the last fetched user info.

jnpsk:


hvge (approved):
Looks OK

PR #790, opened 2025-11-14
Fix #789: Use common shared secret calculation

### LINKED ISSUE #789: Crypto4: Update Shared Secret calculation
### Description

Update shared secret calculation to use common algorithm based on list of KEMs. For EC based algorithms, we're now using DHKEM(P384, HKDF-SHA384) as a replacement for plain ECDH.

### Acceptance criteria

_No response_

### Technical specification

_No response_

### QA specification

_No response_

hvge:
- Use DHKEM for EC based algorithms
- Use NIST approved algorithms for KDF and Key Establishment Scheme (Shared Secret)
- Remove legacy C++ tests, including test vectors


jnpsk (approved):
Looks good!

PR #792, opened 2025-11-20
Fix #791: Crypto4: createJwtSignature response parsing failure

### LINKED ISSUE #791: Crypto4: createJwtSignature response parsing failure

_No response_

### Environment

_No response_

`PowerAuthTestServerAPI#createJwtSignature` fails on the return call because of `[__NSCFString stringValue]: unrecognized selector sent to instance`.

### Log output

_No response_

jnpsk:
As discussed earlier.


hvge (approved):
Thanks for the fix

PR #820, opened 2026-01-27
Android: Support for biometric KEK

### LINKED ISSUE #758: Crypto4: Add support for biometric KEK migration
### Description

In crypto4, we're using 32 Bytes long keys for KEKs protecting authentication code factors, instead of 16 Bytes in V3. This is problematic for Android platform, where user's already prompted for password to perform the protocol upgrade. We don't want to also ask for biometric authentication, so we should extend the existing 16B key to 32Bytes in the upgrade process.

Depends on #752 

### Acceptance criteria

_No response_

### Technical specification

_No response_

### QA specification

_No response_

hvge:
This PR adds functionality that guarantees biometrics to work after upgrade from protocol V3 to V4:

- C++ Session now contains functions for generate KEK. If session is still at protocol V3, then the normalization calculation is compatible with older SDKs (e.g. SHA-256(data), resize to 16B)
- All new biometry related keys are now 256-bits.
- Objective-C wrapper now use KEK functions from C++ session

Fix #758

### DIFF
--- a/include/PowerAuth/Session.h
+++ b/include/PowerAuth/Session.h
@@ -404,12 +404,34 @@ public:
                            const std::string& data_type,
                            SignatureKeyId key_to_use,
                            bool use_compact_form) const;
+
+public:
+    // --------------------------------------------------------------------------------------------
+    // Utilities
+    // --------------------------------------------------------------------------------------------
+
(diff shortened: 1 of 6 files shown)


jnpsk (approved):
Looks good!

PR #824, opened 2026-02-02
Immutable seure vault keys

hvge:
Secure vault keys are not changed after protocol change.

Other changes:

- Android's BaseSdkTest no longer fails


jnpsk (approved):
looks good

PR #827, opened 2026-02-09
Add support for tests with fake biometric key

hvge:
This PR adds support for fake biometric factor key, that allows to test basic biometry-related functionality without actual biometric authentication.

On top of that, there are the following bug-fixes and improvements:
- Android
  - Fixed `CoreCredentials` object construction with biometric factor
  - Fixed saving session's state after authentication code or header calculation
  - Fixed saving session's state after local counter synchronization
  - Expose `CoreSession` object for tests
  - Added more test cases to `AuthenticationCodeTest` class
- iOS
  - Improved test for authentication code calculation, to simulate app restart.


jnpsk (approved):
Looks good

PR #832, opened 2026-02-12
Recovery from failed activation confirm

hvge:
This PR adds ability to recover from failed activation confirmation in protocol 4.0.

Other changes:

- Fixed C++ `Context` cleanup after local activation remove
- `AeadEncryptorFactory` now provides activation scoped key also when activation is not persisted yet. This is required because we need to fetch activation status in confirm activation recovery task.
- Minor improvements in `IActivationService`. Interface now uses references instead of smart pointer copy.
- Added missing getters to status flags in `ActivationStatus`
- Improved error reported from `CoreTask` or `PowerAuthCoreTask`. Previously, most of the errors were "Request failed with no exact reason"

### DIFF
--- a/src/PowerAuth/task/ConfirmActivationTask.cpp
+++ b/src/PowerAuth/task/ConfirmActivationTask.cpp
@@ -67,6 +67,7 @@ void ConfirmActivationTask::onRequestFailure(const Request &request)
     // lock already acquired
     switch (request.getParentTaskTag()) {
         case CONFIRM_ACTIVATION:
+            _session_data->registrationData().v4().lastConfirmFailed = true;
             fetchActivationStatus();
             break;
         case FETCH_STATUS:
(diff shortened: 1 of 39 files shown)


src/PowerAuth/task/ConfirmActivationTask.cpp
```diff
@@ -0,0 +1,121 @@
...
+void ConfirmActivationTask::onTaskStart()
+{
+    LOCK_GUARD();
+    if (_session_data->getCurrentProtocolVersion() < Version_V4) {
+        throw Exception(EC_InternalError, "Confirm task is not supported in V3");
+    }
+    if (!_session_data->hasRegistrationData()) {
+        throw Exception(EC_InternalError, "There's no pending registration to confirm");
+    }
+    if (_session_data->registrationData().v4().lastConfirmFailed) {
```
jnpsk, line 41: Does this flag have any usage? I didn't find code where the `lastConfirmFailed` is set.

hvge: You're right, there's missing set to true in case of failure.

hvge: Fixed in latest commit

jnpsk (approved):
Looks good

hvge:
I'll wait for #835 and then resolve conflicts

PR #839, opened 2026-02-18
Fix #838: Handle session state serialization after upgrade confirm in the ActivationStatusTask

### LINKED ISSUE #838: Handle session state serialization after upgrade confirm in the ActivationStatusTask
### Task specification

It may happen that the protocol upgrade is confirmed during activation status fetching, in which case the `pendingProtocolUpgrade` flag in persistent data is modified. That should lead to saving the session state.
As the session state save must be done explicitly on Android, we need to propagate the session state save requirement to Java from the Task framework.

jnpsk:
It may happen that the protocol upgrade is confirmed during activation status fetching, in which case the `pendingProtocolUpgrade` flag in persistent data is modified. That should lead to saving the session state.

As the session state save must be done explicitly on Android, we need to propagate the session state save requirement to Java from the Task framework.

### DIFF
--- a/proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/impl/GetActivationStatusTask.java
+++ b/proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/impl/GetActivationStatusTask.java
@@ -93,6 +93,8 @@ public class GetActivationStatusTask extends GroupedTask<PowerAuthActivationStat
                     final CoreActivationStatus coreStatus = Objects.requireNonNull(status);
                     if (coreStatus.isSessionSerializationNeeded()) {
                         saveStateCallback.run();
+                    } else if (coreTask.isSessionStateSerializationNeeded()) {
+                        completionListener.onSessionStateChange();
                     }
                     complete(new PowerAuthActivationStatus(coreStatus));
                 }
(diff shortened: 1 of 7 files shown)


proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/impl/GetActivationStatusTask.java
```diff
@@ -93,6 +93,8 @@ public void onNetworkResponse(@Nullable CoreActivationStatus status) {
                     final CoreActivationStatus coreStatus = Objects.requireNonNull(status);
                     if (coreStatus.isSessionSerializationNeeded()) {
                         saveStateCallback.run();
+                    } else if (coreTask.isSessionStateSerializationNeeded()) {
+                        completionListener.onSessionStateChange();
```
hvge: I think this is redundant to `saveStateCallback`. I would remove the method from the completion listener and use `saveStateCallback`. This is my fault that I didn't recognize this when I recently updated the class.

hvge (approved):
I found redundant functionality in the code, otherwise looks OK.

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/impl/GetActivationStatusTask.java
jnpsk: Ok, in that case we can use just `IBiConsumer` interface instead of the custom listener interface.

PR #842, opened 2026-02-19
Fix #836: Add Context to new 2.0 APIs

### LINKED ISSUE #836: Crypto4: Use context in API taking PowerAuthAuthentication
### Task specification

There are several new asynchronous APIs introduced in SDK 2.0.0 on Android that accept a `PowerAuthAuthentication` object as input. Some of them do not take Android’s `Context` object as a parameter, because it is currently not required. We should unify this behavior with the rest of the SDK and make `Context` the first parameter in such methods.

The reason is that we can use `Context` to resolve the KEK for the biometric factor if the application wants to authenticate using biometrics. This change may significantly simplify the use of biometrics on the Android platform.

Examples of such APIs:

- `PowerAuthSDK.calculateJwsSignature()`
- `PowerAuthSDK.calculateDigitalSignature()`

Versus an existing functions:

- `PowerAuthSDK.removeActivationWithAuthentication()`

Note that this task is only about prepare API for the future enhancements. We simply don't want to change API functions in every release of SDK.

### Usage of context explained

In previous SDK versions, we used `Context` to extract the possession or biometric factor every time the factor was needed. In SDK 2.0, the data for the possession factor is extracted during `PowerAuthSDK` initialization, so it is no longer needed later. However, we can still use `Context` to properly handle authentication with biometrics.

hvge:
This PR adds Android `Context` parameter to all new APIs added to `PowerAuthSDK` class. This allows us to improve functionality in the future without another API breaking change. See #836 ticket for more details.

### DIFF
--- a/docs/PowerAuth-SDK-for-Android.md
+++ b/docs/PowerAuth-SDK-for-Android.md
@@ -845,7 +845,7 @@ val signingKey = PowerAuthSignatureKeyId.DEVICE_ML_DSA
 // Data to sign
 val data = "hello".getBytes()
 // Calculate signature
-powerAuthSDK.calculateDigitalSignature(authentication, data, signingKey, object: IDigitalSignatureListener {
+powerAuthSDK.calculateDigitalSignature(context, authentication, data, signingKey, object: IDigitalSignatureListener {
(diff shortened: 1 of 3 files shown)


jnpsk (approved):
👍

PR #846, opened 2026-02-24
Synchronize biometric configuration with server

hvge:
This PR adds feature that automatically synchronize the biometric factor configuration with the server.


jnpsk (approved):
Looks good

PR #852, opened 2026-02-27
Prepare forward data compatibility

hvge:
This PR prepares the SDK for future changes to the local activation data format, similar to the upgrade from protocol V3 to V4.

To support this, we have introduced a new error code (UPGRADE_SDK) and added a function for cleaning up local activation data. This allows applications to remove unsupported data before creating a `PowerAuthSDK` instance.

Other notable changes:

- Added public documentation for missing error codes (both Android and iOS)
- `PA2SessionInterface` now has two step initialization sequence
  - In 1st step, instance is created
  - In 2nd step, initial data load is performed
- `PA2SharedSessionInterface` optimizations
  - There's no longer data re-serialization in the lock function. This is not necessary and in fact, just complicated a lot of things.
- `PA2DefaultSessionInterface` optimizations
  - The initial load doesn't re-serialize session's state. The situation is similar than in PA2SharedSessionInterface implementation
- Removed `PowerAuthErrorCode_CoreError` error code on iOS. That was redundant with "Other" error code

### DIFF
--- a/docs/PowerAuth-SDK-for-iOS.md
+++ b/docs/PowerAuth-SDK-for-iOS.md
@@ -133,9 +133,23 @@ func application(_ application: UIApplication, didFinishLaunchingWithOptions lau
         instanceId: Bundle.main.bundleIdentifier!,
         baseEndpointUrl: "https://<your-domain>/enrollment-server",
         configuration: "ARDDj6EB6iAUtNm...KKEcBxbnH9bMk8Ju3K1wmjbA==")
-
-    // Create a PowerAuthSDK instance with the configuration
-    let powerAuth = try PowerAuthSDK(configuration)
+    do {
+        // Create a PowerAuthSDK instance with the configuration
+        let powerAuth = try PowerAuthSDK(configuration: configuration)
+    } catch let err as NSError {
+        switch err.powerAuthErrorCode {
+        case .wrongParameter:
+            // Invalid configuration.
+        case .invalidActivationData:
+            // Unrecognized data format. You can log this event, clear data and retry SDK construction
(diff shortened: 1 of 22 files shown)


proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/exception/PowerAuthErrorCodes.java
```diff
@@ -172,8 +172,15 @@
      * Digital or JWS signature is not valid.
      */
     int WRONG_SIGNATURE = 25;
+    /**
+     * Upgrade the PowerAuth Mobile SDK in your application. This error may occur if the local
+     * activation data was created with a newer version of the SDK than the one currently used
+     * in your application. This situation can happen if you downgraded your application during
+     * testing.
```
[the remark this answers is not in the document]
hvge: Android doesn't support such feature.

docs/PowerAuth-SDK-for-iOS.md
```diff
@@ -2568,3 +2590,18 @@ private func migrateUserDefaults(appGroup: String) {
     }
 }
 ```
+
+### Upgrade from older SDKs
```
jnpsk, line 2594: This section seems to be omitted from the Android documentation.

jnpsk (approved):

docs/PowerAuth-SDK-for-iOS.md
hvge: That's not necessary, because Android has no activation data sharing feature. We don't have such problem on Android platform.

PR #861, opened 2026-03-04
Fix #844: Biometry key handling for protocol upgrade

### LINKED ISSUE #844: Biometry key handling for protocol upgrade
### Task specification

Followup to https://github.com/wultra/powerauth-mobile-sdk/issues/843

jnpsk:


proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthSDK.java
```diff
@@ -2799,7 +2979,7 @@ public void onBiometricDialogCancelled(boolean userCancel) {
 
             @Override
             public void onBiometricDialogSuccess(@NonNull BiometricKeyData biometricKeyData) {
-                final PowerAuthAuthentication authentication = PowerAuthAuthentication.possessionWithBiometry(biometricKeyData.getDerivedData());
+                final PowerAuthAuthentication authentication = PowerAuthAuthentication.possessionWithBiometry(biometricKeyData.getDerivedData().copy());
                 // TODO: This should be moved in the next release to some global point to make sure that we always clear this object.
                 biometricKeyData.destroy();
```
jnpsk, line 3052: I guess there was missing copy call, otherwise the data are destroyed right away and not accessible from the authentication object

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthSDK.java
```diff
@@ -1370,27 +1370,58 @@ public void removeActivationLocal(@NonNull Context context) {
...
     public @Nullable
     ICancelable startProtocolUpgrade(@NonNull final Context context,
                                      @NonNull final Password password,
-                                     @Nullable final SecureData encryptedBiometryKey,
+                                     @Nullable final PowerAuthAuthentication authentication,
                                      @NonNull final IProtocolUpgradeListener listener) {
 
         try {
-            final CoreTask<CoreProtocolUpgradeResult> task = mSession.startProtocolUpgrade(password, encryptedBiometryKey);
+        if (hasBiometryFactor(context) && !mBiometricConfiguration.isAuthenticateOnBiometricKeySetup() && authentication != null) {
```
hvge: Bad code indentation

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthSDK.java
```diff
@@ -1370,27 +1370,58 @@ public void removeActivationLocal(@NonNull Context context) {
...
+            if (biometricPrompt != null) {
+                // Upgrade is requested for an activation having biometry, authentication on biometry
+                // key setup is not required and biometric prompt is passed. Biometry can be upgraded.
+                return startProtocolUpgradeWithPrompt(context, password, biometricPrompt, listener);
+            }
+        }
+
+        // Only external biometry is upgradable at this point.
+        final boolean hadLocalBiometry = hasBiometryFactor(context);
+        final boolean hadCoreBiometry = mSession.hasBiometryFactor();
```
hvge: The `hasBiometryFactor()` is not that cheep as it looks. I would move this at the beginning of the function and use the result in initial condition.

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthSDK.java
```diff
@@ -2935,6 +3120,55 @@ public void onBiometricDialogFailed(@NonNull PowerAuthErrorException error) {
...
+            final @NonNull IBiometricAuthenticationCallback callback) {
+
+        if (BuildConfig.DEBUG) {
+            // For testing purpose only
+            if (HttpConnectionFailureSimulator.shouldFailAuthenticationUsingBiometrics()) {
+                PowerAuthLog.d("Simulated error on authenticate using biometrics");
+                dispatchCallback(() -> callback.onBiometricDialogFailed(new PowerAuthErrorException(PowerAuthErrorCodes.OTHER, "Simulated error during biometrics authentication")));
+                return new DummyCancelable();
+            }
+        }
```
hvge, line 3216: This is nice. How about to use this function also for an activation persist? We can simulate a similar failure there. It can be done later in separate PR.

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthSDK.java
```diff
@@ -1370,27 +1370,58 @@ public void removeActivationLocal(@NonNull Context context) {
...
+     *                       Or pass authentication object with {@link SecureData} if activation
+     *                       uses external biometry factor key to upgrade biometry key.
      * @param listener A callback with protocol upgrade result.
      * @return {@link ICancelable} associated with the running task.
      */
     public @Nullable
     ICancelable startProtocolUpgrade(@NonNull final Context context,
                                      @NonNull final Password password,
-                                     @Nullable final SecureData encryptedBiometryKey,
+                                     @Nullable final PowerAuthAuthentication authentication,
```
hvge: This is a bit awkward combination of parameters. If we want to pass password and optional biometry in whatever form into the function, then let's use `PowerAuthAuthentication` only, in "persist activation" mode. In this case, we have to call `authentication.validateAuthenticationUsage(true);` in the function to check the purpose of the object.

Other possible solution is:
- Make this function private and replace authentication parameter with two optional parameters:
   ```java
   ICancelable startProtocolUpgrade(@NonNull final Context context,
                                    @NonNull final Password password,
                                    @Nullable final PowerAuthBiometricPrompt biometricPrompt,
                                    @Nullable final SecureData encryptedBiometryKey,
                                    @NonNull final IProtocolUpgradeListener listener)
   ```
- Expose two variants as public API, one taking prompt and another the key data.
- Unfortunately we have to multiple the variants by 2, because the string password variant.

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthSDK.java
hvge: Yes! The bug is there since I introduced `SecureData`, so it's not in production yet. The 1.9.x implementation does a deep copy in that `PowerAuthAuthentication` construction function.

hvge (changes requested):
I suggest a different public API for the upgrade. The combination of Password and PowerAuthAuthentication object is a bit unusual.

proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/PowerAuthSDK.java
```diff
@@ -1366,69 +1366,80 @@ public void removeActivationLocal(@NonNull Context context) {
...
+     * @param biometricPrompt Prompt displayed during the biometric authentication. A "dummy" prompt can be provided.
+     * @param listener A callback with protocol upgrade result.
+     * @return {@link ICancelable} associated with the running task.
+     */
+    public @Nullable
+    ICancelable startProtocolUpgrade(@NonNull final Context context,
+                                     @NonNull final Password password,
+                                     @NonNull final PowerAuthBiometricPrompt biometricPrompt,
+                                     @NonNull final IProtocolUpgradeListener listener) {
+        return startProtocolUpgradeImpl(context, password, biometricPrompt, null, listener);
```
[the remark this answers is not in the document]
jnpsk: The code cannot trigger user authentication, avoiding any UI activity

docs/PowerAuth-SDK-for-Android.md
```diff
@@ -2018,6 +2024,27 @@ val upgradePending = powerAuthSDK.hasPendingProtocolUpgrade()
 
 If this call returns true, the application must perform an activation status fetch to complete the upgrade.
 
+The `startProtocolUpgrade()` method is available in three variants, each
+designed for different biometric configurations.
+
+The variant with `PowerAuthBiometricPrompt` allows the biometric key to be
+upgraded during the protocol upgrade, if the SDK configuration has the
+`authenticateOnBiometricKeySetup` disabled or external biometric key is not
+used. If the SDK configuration has the `authenticateOnBiometricKeySetup`
```
hvge: I think it would be nice to provide an example, or at least mention that `.noPromptForBiometricKeySetup(context)` should be provided.

hvge (approved):
Looks OK. 

The documentation is formally OK, but I'm afraid that we'll have to provide a more complex examples for the most common scenarios (usage of AES, or RSA key for biometric factor).

PR #867, opened 2026-03-09
Fixed failing tests on iOS

hvge:
This PR is composed of multiple related commits and fixes the following issues:
- Share information about External Pending Operation on iOS, when an Activation Data Sharing feature is enabled.
- Failing tests related to Activation Data Sharing on iOS.

## Fix for External Pending Operation (EPO)
- Fixed missing EPO start with protocol upgrade type
- `ProtocolUpgradeTask` is now started immediately after its creation. This prevents premature end of the EPO immediately after the task creation.
- Added ability to reclaim EPO created in the same application if PowerAuthSDK instance is re-created and the shared data still contains information about EPO for upgrade.

## Integration tests fix
- Added proper simulation of app restart by force release internal PowerAuthSDK instance resources.

## Other changes:
- Added debug description to `PowerAuthActivationStatus` and `PowerAuthExternalPendingOperation`
- Fixes integration tests for tvOS platform


jnpsk (approved):
Looks OK

PR #872, opened 2026-03-13
Fix #782: Crypto4: Better error report when SDK config doesn't match algorithm

### LINKED ISSUE #782: Crypto4: Better error report when SDK config doesn't match algorithm
### Description

If SDK configuration string provided to SDK doesn't contain keys required for the selected algorithm, then the generic error is reported back to the application. This should be improved because I feel it will be a source of a lot of supporting ticket in the future.

We should report at least information that SDK configuration is insufficient for the selected algorithm.

### Acceptance criteria

_No response_

### Technical specification

_No response_

### QA specification

_No response_

jnpsk:


jnpsk:
Intention of this PR was to avoid reporting via exception and keep the behaviour of public APIs, that is not required and we can implement throwing alternatives of configuration validation methods, in which case PR https://github.com/wultra/powerauth-mobile-sdk/pull/871 solves that better.

PR #895, opened 2026-05-19
Fixed reset of C++ Session

hvge:
This PR fixes reset of  C++ `Session` object in case the current protocol specification is different than specification from the configuration. The core of the problem was in caching the specification in `Context` and not rollback the value to the target specification, required by the application.


jnpsk (approved):
Thanks!

PR #901, opened 2026-06-01
Note about 1FA activation removal

hvge:
This PR adds a note about possibility to remove the activation with using 1FA authentication.


jnpsk (approved):
Looks good

PR #902, opened 2026-06-02
Fix for broken Android biometry

hvge:
This PR updates library dependencies, most notably `androidx.biometric:biometric:1.4.0-alpha07`. The primary purpose of this update is to avoid possible compatibility issues if the application is using `androidx.core:core-ktx:1.18.0` and newer.

As a side effect of this change, the minSdk is raised to 23 and therefore the change also removes dead code for no longer supported Android versions.

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -10,7 +10,7 @@
 
 ### Both platforms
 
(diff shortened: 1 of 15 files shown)


jnpsk (approved):
Looks good!

PR #905, opened 2026-06-02
Support for temporary activation block

hvge:
This PR adds support for the temporary activation block, introduced in PowerAuth Server 2.2. The feature allows automatic activation unblock after a period of time. To support such a feature on the mobile side, the SDK now provides information about when the activation will be unblocked.

Other changes:
- Integration Tests now supports PowerAuth Server 2.2


jnpsk (approved):
Nice!

PR #912, opened 2026-07-02
Fixed system prompt when upgrading biometry on iOS

### LINKED ISSUE #909: FaceID required during `startProtocolUpgrade` on iOS

2.0.0 RC1

### Environment

_No response_

When `startProtocolUpgrade` is run on iOS, it automatically prompts FaceID during the process.

The trigger is launched here, based on my debugging:

https://github.com/wultra/powerauth-mobile-sdk/blob/develop/proj-xcode/PowerAuth2/PowerAuthSDK.m#L1020

Tested on the latest iOS 26.5.1

### Log output

_No response_

kober32:
Fixes #909

The issue was that the upgrade was updating the value (which requires biometric approval). The new approach is removing and reintroducing the biometry again.

The same approach is used in `persistActivation`.


hvge (approved):
Thank you!

PR #918, opened 2026-07-08
Context::resetState fix

### LINKED ISSUE #917: `removeActivationLocal` does not work when protocol upgrade is available

2.0.0 BETA, tested on iOS

### Environment

_No response_

If `removeActivationLocal` is called when protocol upgrade is available, it does not clear the data.

### Log output

_No response_

kober32:
Fixed #917: "resetState" bug.

When there was a pending upgrade, the activation could not be removed locally because the data were never cleaned, and services were just recreated.

I also added tests to cover it. When you remove the change in the Context class, the test fails...


jnpsk (approved):
Looks good, thanks for the additional tests!

PR #920, opened 2026-07-17
Fix #910: Fix missing class lombok.Generated issue

### LINKED ISSUE #910: Fix missing class `lombok.Generated` issue

_No response_

### Environment

_No response_

As I am implementing PQA on mToken Android, I have stumbled upon build error:
```
ERROR: Missing classes detected while running R8. 
> Task :app:minify*****DevReleaseWithR8
ERROR: R8: Missing class lombok.Generated (referenced from: boolean io.getlime.core.rest.model.base.entity.Error.canEqual(java.lang.Object) and 15 other contexts)
```

I am now aware of introducing Lombok dependency in mToken codebase, it looks (guess) that there are some REST annotation maybe from Java backend classes? 
Workaround is to silence R8 warnings, but we need to fix this before distributing to our clients.
Thanks :)

### Log output

_No response_

jnpsk:
Fixes https://github.com/wultra/powerauth-mobile-sdk/issues/910

The described error originates from the `rest-model-base` dependency, specifically the `Error` class, which uses Lombok to generate boilerplate methods. Lombok then annotates its generated methods with `@lombok.Generated`. During the build, R8 attempts to resolve this annotation but fails, since the Lombok library itself was never added as a dependency.

Since `lombok.Generated` is not runtime annotation it seems safe to just suppress the unresolved reference. Alternatively, we have to add Lombok as a compile dependency, which would let R8 resolve the class properly, but suppressing the specific warning is simpler and has no functional downside here.

### DIFF
--- a/proj-android/PowerAuthLibrary/proguard-rules.pro
+++ b/proj-android/PowerAuthLibrary/proguard-rules.pro
@@ -30,3 +30,4 @@
 }
 # necessary for R8 fullMode
 -keep, allowobfuscation class io.getlime.core.rest.model.base.**
+-dontwarn lombok.Generated


marek-ch (approved):

kober32 (approved):
ok

PR #924, opened 2026-07-29
Fix #922: Low level exception of authentication header creation methods on missing activation

### LINKED ISSUE #922: legacy: `authenticationHeaderForRequestWithParams/Body` throw "PersistentData not available" without activation

2.0.0

### Environment

_No response_

found in flutter migration, this is an example ios code used. This only happens in legacy algorithm 
```swift
private func authenticationHeaderForRequestWithBody(_ call: FlutterMethodCall, _ result: @escaping FlutterResult) throws {
        try usePowerAuth(call, result) { sdk, _ in
            let auth = try constructAuthentication(call)
            let uriId: String = try call.requireParameter(Args.uriId)
            let method: String = try call.requireParameter(Args.method)
            let data = call.optionalDataParameter(Args.body)
            let header = try sdk.authenticationHeaderForRequestWithBody(
                with: auth,
                method: method,
                uriId: uriId,
                body: data
            )
            result(header.serializable)
        }
    }
```

### Log output

_No response_

jnpsk:


kober32 (approved):
👍

PR #937, opened 2026-09-11
Fix #936: Automatic time synchronization no longer guaranteed when obtaining an E2EE encryptor

### LINKED ISSUE #936: Automatic time synchronization no longer guaranteed when obtaining an E2EE encryptor
### Summary

In 1.9.x, obtaining an ECIES encryptor via `getEciesEncryptorFor*Scope()` guaranteed the SDK time was synchronized with the server, and `encryptRequest` was tolerant of unsynchronized time. In 2.0.x this guarantee is gone: if a temporary key is already cached, obtaining an encryptor performs no network call and no time sync, and `encryptRequest` now throws a hard `EC_TimeNotSynchronized`. Because time synchronization is reset on foreground transitions while the cached key is retained, apps performing custom E2EE encryption can hit `EC_TimeNotSynchronized` in a scenario that silently worked in 1.9.x.

This contradicts the documented contract (docs still list "*After receiving any response encrypted with our End-To-End Encryption scheme*" as an automatic sync trigger, and the 1.8→1.9 migration guide states the async `getEciesEncryptorFor*Scope()` methods synchronize time automatically).

### Where the sync happens

**1.9.x (tag `1.9.6`) — sync guaranteed when getting an encryptor; encrypt is non-fatal**

Getting the encryptor fetches the temporary key, and the key fetch drives the sync. Even without sync, `encryptRequest` only warned and used local time:

- iOS key fetch → sync: [`PA2KeystoreService.m#L102`](https://github.com/wultra/powerauth-mobile-sdk/blob/1.9.6/proj-xcode/PowerAuth2/private/PA2KeystoreService.m#L102) (start) and [`#L156`](https://github.com/wultra/powerauth-mobile-sdk/blob/1.9.6/proj-xcode/PowerAuth2/private/PA2KeystoreService.m#L156) (complete)
- iOS encrypt = warn only: [`PowerAuthCoreEciesEncryptor.mm#L115-L119`](https://github.com/wultra/powerauth-mobile-sdk/blob/1.9.6/proj-xcode/PowerAuthCore/PowerAuthCoreEciesEncryptor.mm#L115-L119)
- Android key fetch → sync: [`DefaultKeystoreService.java#L116`](https://github.com/wultra/powerauth-mobile-sdk/blob/1.9.6/proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/impl/DefaultKeystoreService.java#L116) & [`#L143`](https://github.com/wultra/powerauth-mobile-sdk/blob/1.9.6/proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/sdk/impl/DefaultKeystoreService.java#L143)
- Android encrypt = warn only: [`EciesEncryptor.java#L198-L202`](https://github.com/wultra/powerauth-mobile-sdk/blob/1.9.6/proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/core/EciesEncryptor.java#L198-L202)

**2.0.x (tag `2.0.0`) — sync only if the key fetch fires; encrypt now throws hard**

- Sync is now a side-effect of the key fetch, and is skipped when the key is cached: [`PA2CoreHttpClient.m#L146-L147`](https://github.com/wultra/powerauth-mobile-sdk/blob/2.0.0/proj-xcode/PowerAuth2Private/PA2CoreHttpClient.m#L146-L147); core sync in [`EciesEncryptorFactory.cpp#L194`](https://github.com/wultra/powerauth-mobile-sdk/blob/2.0.0/src/PowerAuth/v3/EciesEncryptorFactory.cpp#L194) & [`#L260`](https://github.com/wultra/powerauth-mobile-sdk/blob/2.0.0/src/PowerAuth/v3/EciesEncryptorFactory.cpp#L260)
- Cached-key short-circuit (no network, no sync): [`PA2KeystoreService.m#L133-L137`](https://github.com/wultra/powerauth-mobile-sdk/blob/2.0.0/proj-xcode/PowerAuth2Private/PA2KeystoreService.m#L133)
- The breaking throw (ECIES v3): [`EciesEncryptor.cpp#L180-L181`](https://github.com/wultra/powerauth-mobile-sdk/blob/2.0.0/src/PowerAuth/v3/EciesEncryptor.cpp#L180-L181)
- The breaking throw (AEAD v4): [`AeadEncryptor.cpp#L78-L79`](https://github.com/wultra/powerauth-mobile-sdk/blob/2.0.0/src/PowerAuth/v4/AeadEncryptor.cpp#L78-L79)

### Steps to reproduce

1. Create an activation and let the SDK synchronize time (e.g. via activation status).
2. Obtain an application- or activation-scoped encryptor so a temporary key gets cached.
3. Send the app to background and bring it back to foreground (this resets time synchronization but keeps the cached temporary key).
4. Obtain an encryptor again via `getEncryptorFor*Scope()` — it returns immediately with no network call and no time sync.
5. Call `encryptRequest`.

### Expected

Obtaining an encryptor guarantees synchronized time (as in 1.9.x and as documented), so encryption succeeds.

### Actual

`encryptRequest` throws `EC_TimeNotSynchronized` (iOS `NSError` code `PowerAuthErrorCode_TimeSynchronization` / Android `PowerAuthErrorException` `TIME_SYNCHRONIZATION`).

### Affected versions

2.0.0 (introduced by the protocol-v4 core rework). Both iOS and Android.

### Additional note

While tracing this, I noticed a likely related latent issue: the ECIES v3 encryptor's `_fail_on_nosync_time` member is not initialized in the constructor used in production ([`EciesEncryptor.cpp#L138`](https://github.com/wultra/powerauth-mobile-sdk/blob/2.0.0/src/PowerAuth/v3/EciesEncryptor.cpp#L138)) and has no default in the header ([`EciesEncryptor.h#L54`](https://github.com/wultra/powerauth-mobile-sdk/blob/2.0.0/src/PowerAuth/v3/EciesEncryptor.h#L54)), so its value is indeterminate. The AEAD v4 encryptor always sets it to `true`.

jnpsk:
Fixes https://github.com/wultra/powerauth-mobile-sdk/issues/936

End-to-end request encryption no longer fails immediately when local time is not synchronized. The SDK now logs a warning and continues with request encryption.

### DIFF
--- a/docs/Changelog.md
+++ b/docs/Changelog.md
@@ -1,7 +1,14 @@
 # Changelog
 
 <!--------------------------------------------------->
-## 2.0.0 (TBA - Not released yet)
+## TBA - Not released yet
+
+### Both platforms
+
+- End-to-end request encryption no longer fails immediately when local time is not synchronized. The SDK now logs a warning and continues with request encryption ([936](https://github.com/wultra/powerauth-mobile-sdk/issues/936)).
+
+<!--------------------------------------------------->
+## 2.0.0 (August 2026)
 
 ### Important notice
 
--- a/proj-android/PowerAuthLibrary/src/androidTest/java/io/getlime/security/powerauth/integration/tests/BaseSdkTest.java
+++ b/proj-android/PowerAuthLibrary/src/androidTest/java/io/getlime/security/powerauth/integration/tests/BaseSdkTest.java
@@ -19,6 +19,7 @@ package io.getlime.security.powerauth.integration.tests;
 import androidx.annotation.NonNull;
 
 import io.getlime.security.powerauth.core.CoreEncryptor;
+import io.getlime.security.powerauth.core.CoreEncryptedRequest;
 import io.getlime.security.powerauth.core.SecureData;
 import io.getlime.security.powerauth.keychain.Keychain;
 import io.getlime.security.powerauth.keychain.KeychainFactory;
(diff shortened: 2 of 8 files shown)


kober32 (approved):
LGTM. 

I also think that we should improve this behavior by specifically request time sync along with the warning. Could you create an issue with some high-level proposed solution?

PR #942, opened 2026-09-16
Fix #941: Revert activation status enum ordinal values

### LINKED ISSUE #941: Undocumented `PowerAuthActivationState` change
## The issue

The migration guide to `2.0.0` mentions that the `Created` state was removed from Android and iOS.

But other than that, it also changed what `int` values these enums are mapped to:

| Version | Android | iOS |
|---------|---------|-----|
| 1.9.5 | [ActivationStatus.java@1.9.5](https://github.com/wultra/powerauth-mobile-sdk/blob/1.9.5/proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/core/ActivationStatus.java#L32-L55)| [PowerAuthActivationStatus.h@2.0.0](https://github.com/wultra/powerauth-mobile-sdk/blob/1.9.5/proj-xcode/PowerAuth2/PowerAuthActivationStatus.h#L24-L49) |
| 2.0.0 | [CoreActivationState.java@2.0.0](https://github.com/wultra/powerauth-mobile-sdk/blob/2.0.0/proj-android/PowerAuthLibrary/src/main/java/io/getlime/security/powerauth/core/CoreActivationState.java#L33-L54) | [PowerAuthActivationStatus.h@2.0.0](https://github.com/wultra/powerauth-mobile-sdk/blob/2.0.0/proj-xcode/PowerAuth2/PowerAuthActivationStatus.h#L24-L49) |

## Why is it problem

For example, the state `active` was changed from `3 -> 1`.

This change is not documented anywhere and is breaking, especially if our customers use these values to process or cache the status based on it.

## Solution

We should consider if this change is acutaly wanted and:
1. document it in the migration guide
2. revert it back to keep the values the same

jnpsk:
Fix https://github.com/wultra/powerauth-mobile-sdk/issues/941 by restoring activation status enum ordinal values to maintain compatibility with PowerAuth SDK 1.9.x The values are in sync across C++, iOS, and Android.


kober32 (approved):
👍

